PR #4, opened 2015-11-03
Build fixes

tombriden:
I had to make these changes to get a working AOSP marshmallow build including opengapps.

### DIFF
--- a/opengapps-packages.mk
+++ b/opengapps-packages.mk
@@ -1,3 +1,4 @@
+include vendor/google/build/config.mk
 include $(GAPPS_FILES)
 
 PRODUCT_PACKAGES += GoogleBackupTransport \
--- a/README.md
+++ b/README.md
@@ -23,13 +23,13 @@ and add the following towards the end:
 
 In your `device/manufacturer/product/BoardConfig.mk` file, towards the end, add:
 ```makefile
-OPENGAPPS_VARIANT := <variant>
+GAPPS_VARIANT := <variant>
 ```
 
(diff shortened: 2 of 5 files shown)


Blystad:
Hey,

Thank you for your contribution.

Sorry for not attending to your pull request. I forgot to set the correct "Watch" setting on Github.

I'll test your Pull Request tomorrow, and get back to you.

opengapps-packages.mk
```diff
@@ -1,3 +1,4 @@
+include vendor/google/build/config.mk
```
Blystad: I can't see why this is needed, this should get included by the existing AOSP build system, even on M: https://android.googlesource.com/platform/build/+/android-6.0.0_r26/core/main.mk#106

mfonville: I have literally 0% experience with these .mk files, but the line you reference includes a minus-sign before the include, could that have any effect?

Blystad: The minus before just means "if we're unable to find the file, ignore it
and continue"

Since we have a file at `vendor/google/config.mk` then that should be
included very early in the build process.

mfonville: Ok, thanks for the explanation. Then the behaviour is indeed weird.

tombriden: Hi, I agree it looks like it should work without that line; I just double checked by removing it and nothing from opengapps-files gets included in the build and adding the following just prints an empty string
```
$(warning $(GAPPS_FILES))
```

Blystad: Seems you're right. Good catch.

Blystad:
Tested here now, LGTM.

PR #5, opened 2015-11-11
modules: Remove "Auto-generated makefile text"

Blystad:
This was an artifact from when I originally created the makefiles automatically.


Blystad:
I'll merge this in ~24h unless there's any objections.

PR #6, opened 2015-11-11
build: Require GAPPS_VARIANT to be set

Blystad:
I just had a build issue because I was using too much space on the system partition,
since it set GAPPS_VARIANT to default "stock" when opengapps-packages.mk got imported
by the device.mk

This avoids this happening by requiring that GAPPS_VARIANT is set.

Also fix so we don't reassign a set variable (GAPPS_VARIANT) but rather move it
to a seperate variable (TARGET_GAPPS_VARIANT) that is being read.

Move the configuration variable recommendation from BoardConfig.mk -> device.mk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ and add the following towards the end:
 
 **2. Set the desired OpenGapps variant**
 
-In your `device/manufacturer/product/BoardConfig.mk` file, towards the end, add:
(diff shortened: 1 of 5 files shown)


Blystad:
ref #4 - the build_fixes made everything at my end default to "Stock" instead of what I had configured ("Micro")

Blystad:
I'll merge this in ~24h unless there's any objections.

PR #8, opened 2015-12-19
new package: com.google.android.webview

Blystad:
Added webview based on PR #1 from @sevenrock and set it to the correct default package, with a AOSP framework resource override.

### DIFF
--- a/modules/WebViewGoogle/Android.mk
+++ b/modules/WebViewGoogle/Android.mk
@@ -1,0 +1,7 @@
+LOCAL_PATH := .
+include $(CLEAR_VARS)
+LOCAL_MODULE := WebViewGoogle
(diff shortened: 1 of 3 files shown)


mfonville:
:+1:

mstfkaratas:
Does "LOCAL_MODULE_OVERRIDES := webview" work? because webview folder and corresponding lib are still in system/app and system/lib after successfull build of cm12.1.

Blystad:
Blergh, I misspelled it...it's actually

LOCAL_OVERRIDES_PACKAGES

I also need to move the webview package up to stock. I considered making it
a core package but to ensure the opengapps matrix matches we should put it
under stock.

Blystad:
Fixed the last issues now.

PR #14, opened 2016-01-16
Fix #9 link lib from api levels as needed

### LINKED ISSUE #9: Could not solve api level
Hi all im triying build them into custom device, my surprise are that libraries are all into subdir and these could not be solved by api lvl when lauch build sistem try to copy libs from sources/arm/lib/(lib_name) but rigth path is under sources/arm/lib/(api level)/(libname), can some one check these issue thanks, regards.

jcppkkk:
Llink lib to different api level as new as possible


mfonville:
Note that lib folders can also be named 'lib64' instead of 'lib' (on the 64 bit platforms)

jcppkkk:
@mfonville It seems rules in [AndroidProducts.mk](https://github.com/opengapps/aosp_build/blob/master/modules/AndroidProducts.mk) only handle arm/lib now. 

It could be another To-do task to handle if different targets (e.g. arm64) are added into .repo/manifests/

Blystad:
I'm closing this. AndroidProducts.mk shouldn't have been included in the first place. `opengapps-files.mk` is whats used for shared files, and libraries should be referenced in the application that needs them.

Also see #14

PR #15, opened 2016-01-16
Refactor files.mk to avoid manually specifying all files

Blystad:
This refactors opengapps-files.mk to avoid specifying every file, replacing
it instead with two functions that search and find all the necessary files
to copy over.

Also remove AndroidProducts.mk since that file is not supposed to be used.
It's a remnant of the original code generation solution.

Related to #9 - superceeds #14


opengapps-files.mk
```diff
@@ -1,66 +1,7 @@
...
-                      $(GAPPS_SOURCES_ALL_PATH)/etc/permissions/com.google.android.maps.xml:system/etc/permissions/com.google.android.maps.xml \
-                      $(GAPPS_SOURCES_ALL_PATH)/etc/permissions/com.google.android.media.effects.xml:system/etc/permissions/com.google.android.media.effects.xml \
-                      $(GAPPS_SOURCES_ALL_PATH)/etc/preferred-apps/google.xml:system/etc/preferred-apps/google.xml \
-                      $(GAPPS_SOURCES_ALL_PATH)/framework/com.google.widevine.software.drm.jar:system/framework/com.google.widevine.software.drm.jar \
-                      $(GAPPS_SOURCES_ALL_PATH)/framework/com.google.android.maps.jar:system/framework/com.google.android.maps.jar \
-                      $(GAPPS_SOURCES_ALL_PATH)/framework/com.google.android.dialer.support.jar:system/framework/com.google.android.dialer.support.jar \
-                      $(GAPPS_SOURCES_ALL_PATH)/framework/com.google.android.media.effects.jar:system/framework/com.google.android.media.effects.jar
+PRODUCT_COPY_FILES += $(call gapps-copy-to-system, all, etc/permissions etc/preffered-apps framework)
```
Blystad: Spelling error, should be `preferred-apps`.

@mfonville should we add `sysconfig/google.xml` to MM+? (always) and `sysconfig/google_build.xml` to MM+ for `stock` builds?

mfonville:
I think both are safe to all MM+, independent of package size.
There is no disadvantage for having the google_build.xml there I think and keeps it simple and equal to the flashable package (there we aways include everything from /etc/)

Blystad:
Updated to just copy the entire etc/ dir...for lazyness. 

I've now also fixed this for the new library structure.

Going to do a clean build here and test it.

Blystad:
Clean build passed and everything looks great. Merging this.

PR #24, opened 2016-03-12
New packages, overrides and fixes.

Dudulle:


opengapps-packages.mk
```diff
@@ -12,22 +12,28 @@ PRODUCT_PACKAGES += GoogleBackupTransport \
                     SetupWizard \
                     Phonesky \
                     GoogleCalendarSyncAdapter
+                    
+ifneq ($(filter $(call get-allowed-api-levels),23),)
```
Blystad: Shouldn't GoogleTTS override for LP as well?

opengapps-packages.mk
```diff
@@ -12,22 +12,28 @@ PRODUCT_PACKAGES += GoogleBackupTransport \
...
 
 ifneq ($(filter $(TARGET_GAPPS_VARIANT),micro),) # require at least micro
 PRODUCT_PACKAGES += CalendarGooglePrebuilt \
                     PrebuiltExchange3Google \
                     PrebuiltGmail \
-                    GoogleHome \
-                    GoogleTTS
+                    GoogleHome
+                    
+ifeq ($(filter $(call get-allowed-api-levels),23),)
```
Blystad: and why is there a duplicate here?

opengapps-packages.mk
mfonville: GoogleTTS was before 6.0 optional, since 6.0 it is required for SetupWizard

opengapps-packages.mk
Dudulle: This one adds tts if the API level is below 23 so it is not added twice on Marshmallow.

Blystad:
Fair enough. Thanks for your contribution :+1:

PR #25, opened 2016-03-12
Add dpi selection

Dudulle:


mfonville:
Very nice PR. An addendum from me: the mechanism provided in this code puts DPI as 1st selection priority, so if a DPI version is e.g. not the very latest version it uses it nonetheless. This is opposite behaviour of the OpenGApps installable-package that prefers newer versions over matching DPI.
This behaviour is consistent though with what is expected so +1,  but it would be a nice feature for the future to be able to switch between the two behaviours.

Dudulle:
Thanks for your feedback.
I will look into this next week if I have enough time.

Meanwhile, this code also selects the nodpi package if it has a higher version than the selected dpi. So in the worse case it will not install an older version it would have without this change.

Dudulle:
I have adjusted the code to handle the latest packages.
I will test it with a clean build later and push the commits if it works as expected.

Dudulle:
This works but requires the aapt command.
It is built when building AOSP but it is required too early (when including the .mk files).
if it is not installed and we do a clean build, this will fail. Installing/Building aapt first or building again after it failed will fix the problem.

mfonville:
ok. No we will have to wait for @Blystad to review and merge it :)

Blystad:
Hey, there is an aapt command in `prebuilts/sdk/tools/linux/aapt`, try using that :)

https://android.googlesource.com/platform/prebuilt/+/master/sdk/tools/linux/

Dudulle:
Right, I saw it but I guess I was too obsessed with the one being compiled.
I will make the necessary changes.

Blystad:
Awesome. Other then that it LGTM. So when you've made the change I'll merge it.

Dudulle:
Thanks.
I've updated the last commit.

Blystad:
Perfect. Thanks for your contribution.

PR #27, opened 2016-03-18
Fix WebView dependencies.

Dudulle:
WebView requires libwebviewchromium_loader and libwebviewchromium_plat_support libraries.

When overriding aosp WebView, those librares are missing.
This causes most gapps and other apps using the WebView to crash.
Adding them to LOCAL_REQUIRED_MODULES will fix this issue.


Blystad:
Nice find. Thanks for the contribution.

PR #28, opened 2016-03-19
Add helper file for package overrides.

Dudulle:
~~This file has to be included in each modules which requires those features.~~
This ~~file~~ adds the following features:
-Check for minimum OpenGApps variant (use GAPPS_LOCAL_OVERRIDES_MIN_VARIANT := [variant] instead of condition)
-New global variable to force overrides (GAPPS_FORCE_PACKAGE_OVERRIDES := true)
-Force overrides on per-app basis (GAPPS_PACKAGE_OVERRIDES += [GAppsModule])
-Bypass overrides if allowed (GAPPS_BYPASS_PACKAGE_OVERRIDES += [GAppsModule])


Blystad:
Hey,

Cool idea. With a little bit of refactoring we should be able to improve this a bit.

The standard AOSP build system thinks of including files as either build targets or for target clean-up. So I'm not a fan of having to manually import a non-build-target file into the various modules, and we could get away with not doing it.

I also don't entirely understand the use of `GAPPS_BYPASS_PACKAGE_OVERRIDES`.

Anyhow, here's how I suggest you handle it: 

in `core/prebuilt_apk.mk` check for the presence of `GAPPS_OVERRIDES_MIN_VARIANT`, and `GAPPS_PACKAGE_OVERRIDES`. 
If they fulfill the conditions you have today, _then_ assign the `LOCAL_OVERRIDES_PACKAGES` variable, 
and clean-up `GAPPS_OVERRIDES_MIN_VARIANT` and `GAPPS_PACKAGE_OVERRIDES` afterwards.

Dudulle:
Hi,

Thanks for your feedback.

The reason I have placed this in a separate file is so we can exclude this checks for the modules that do not need this. 
This also allows to set the local overrides after this file so they cannot be removed.
Using a new local variable to store the overrides would also handle this and checking if that variable is empty avoids making too much checks for nothing so I don't mind changing it.

To be a little bit more clear, I defined GAPPS_BYPASS_PACKAGE_OVERRIDES and GAPPS_PACKAGE_OVERRIDES as global variables so we can enable/disable the overrides from the device.mk file on a per-app basis.
Using the BYPASS variable can be useful if we want to keep a stock package with a variant which overrides it or with the FORCE_OVERRIDES flag. I do not use it but I thought it could be useful if ever there is a problem on some builds when a stock package is missing.

I'll push an update in a few minutes.

Blystad:
I think the updated version looks great. Thanks for the contribution.

sevenrock:
Hi, this PR breaks compilation, reverting it solves the problem:

============================================
PLATFORM_VERSION_CODENAME=REL
PLATFORM_VERSION=6.0.1
CM_VERSION=13.0-20160323-UNOFFICIAL-peregrine
TARGET_PRODUCT=cm_peregrine
TARGET_BUILD_VARIANT=userdebug
TARGET_BUILD_TYPE=release
TARGET_BUILD_APPS=
TARGET_ARCH=arm
TARGET_ARCH_VARIANT=armv7-a-neon
TARGET_CPU_VARIANT=krait
TARGET_2ND_ARCH=
TARGET_2ND_ARCH_VARIANT=
TARGET_2ND_CPU_VARIANT=
HOST_ARCH=x86_64
HOST_OS=linux
HOST_OS_EXTRA=Linux-4.5.0-rc7-amd64-x86_64-with-debian-stretch-sid
HOST_BUILD_TYPE=release
BUILD_ID=MHC19J
OUT_DIR=/home/android/cm13.0/out
============================================
vendor/google/build/core/clear_vars.mk:1: *** Too many open files.

Blystad:
Hmm, maybe the addition of the new custom clear_vars file puts you just over the limit.

Can you try raising your limit of open files? See: https://easyengine.io/tutorials/linux/increase-open-files-limit/

sevenrock:
I will try later in the day, for now the current limits:

$ cat /proc/sys/fs/file-max
3254851
$ ulimit -Hn
65536
$ ulimit -Sn
65536

Dudulle:
@niks255 can you try to raise the max number of open files as explained by @Blystad ?
By the way, are you using pure AOSP or another ROM?

niks255:
I inreased the limit to 2097152 and no luck.
I'm building CM13, not AOSP.

Dudulle:
Did you raise the limit of user files too or only system?
You might need to reboot your computer too in order for the changes to take effect.

niks255:
Seems like reboot solved the issue.
No I have another problem. I can't build it, it throws error right after I run breakfast %devicename%. Error tells me it's vendor/google/build/config.mk line 29, but I do have GAPPS_VARIANT set.

Dudulle:
OK good.
GAPPS_VARIANT needs to be defined in your device's device.mk file.
You also need to include the opengapps-packages.mk file at the end of your device.mk file.
If you set it in BoardConfig, you will get this error

niks255:
Ah, ok. Seems like it works.

sevenrock:
Raising ulimit let the "Too many open files" error disappear, but instead the make process takes 100% cpu load on one of my cpu (i7-3770S) cores and loops for over half an hour, I cancelled make then. I guess something is still wrong here (32gb ram, Linux debian 4.5.0-rc7-amd64 #1 SMP Debian 4.5~rc7-1~exp1 (2016-03-07) x86_64 GNU/Linux).

Dudulle:
This is normal. Compiling Android takes a lot of resources.
I have a last gen i7 and the build takes me about 90 minutes using "make -j12" (12 threads). I you use only one thread it will take even longer.

sevenrock:
Sorry, after reading my text again I understood that I made myself not clear :) After starting make the info text I quoted appears, the last consisting of ===================. Here the screen output stops and make seems to be caught in an endless loop, no screen output for > 30 minutes. Normally the build begins a few seconds after the ========= line. It has been like that for several years now :)

Dudulle:
Yes that should not happen.
Could you show the configuration you are using (everything related to opengapps in device.mk)

niks255:
I increased all the limits to a very high value but it still fails to build.

Dudulle:
@niks255 Your build wont work because of the changes you've made.
You are removing variables which are required.
Do not change the code and set up the configuration you want in your device.mk file.

niks255:
@Dudulle Look at my local manifest for google apps https://github.com/niks255/local_manifests/blob/cm-13.0/opengapps.xml
Now what is it you're talking about?

Dudulle:
This: https://github.com/niks255/aosp_build/commit/ddfc8bf32093f41f456ebc644e8218a5b47d19fd

niks255:
@Dudulle But it's not what I'm using. You can see it in local manifest.

Dudulle:
Your local manifest will not override changes you've made. You need to checkout with git and configure your device.mk as explained in the readme file.

niks255:
@Dudulle Damn, what makes you think I don't know that? I did everything that was necessary and added everything that was needed https://github.com/niks255/android_device_semc_msm7x30-common/commit/85bae2c80141720e16a9e1e5208d57f4534a6014
The problem is still the same:
vendor/google/build/core/clear_vars.mk:1: *** Too many open files

Dudulle:
Can you check again if your config.mk is exactly as it is in this repository?
I will test your configuration here when I can and see if it reproduces the same issue.

niks255:
@Dudulle Here's the command
image
Here's the output
image

Dudulle:
Can you see if this commit fixes the error?
https://github.com/Dudulle/aosp_build/commit/e5d19d79c04700c93fa321afb81bce5254784ccf

Dudulle:
PR #31 might also improve this issue.

niks255:
@Dudulle Where's your commit? I can't cherry-pick it.

Dudulle:
You must fetch it first with
`git fetch https://github.com/Dudulle/aosp_build.git e5d19d79c04700c93fa321afb81bce5254784ccf`

niks255:
Yeah, I know. Didn't realize it was in another branch.

Dudulle:
OK thanks.
You used only this commit or you had to merge PR #31 too?

niks255:
@Dudulle I uploaded my current source https://github.com/niks255/aosp_build

Dudulle:
OK,
PR #32 should be enough then.
Thanks

niks255:
@Dudulle BTW there was an issue with google keyboard when it was not visible in the google play after flashing google apps. I mean, Google Play just didn't list it as installed. Is it fixed now?

Dudulle:
This issue doesn't seem to be present using this project.
I see Google keyboard as installed in Play Store

niks255:
Ah, ok then.

Dudulle:
This is for the "OK Google" voice command.

niks255:
@Dudulle It does work for me without it. Or is it for calling from any screen, not just google now app?

Dudulle:
I think this is for any screen.
Not sure this is required if you use both Google app and Google Home.

mfonville:
@Dudulle @niks255 It is for "Ok Google"-support even if the screen is off. Only a few devices (N9, N6, N6P, N5X) support this, probably dependent on certain hardware and very probably ROMs will need some tweaks too to make it work.

niks255:
@mfonville Ah, now I got it. AFAIK these devices have a separate low-power cpu that never goes into the deep sleep. At least first moto devices with voice wake up had it.

sevenrock:
@Dudulle: PR #33 fixed the problem, thanks!

PR #30, opened 2016-03-20
Use AOSP local_manifests feature

fonic:
Instead of altering existing manifests, additional 'overlay' manifests can be but in `.repo/local_manifests`. The repo tool automatically includes files in this folder.

[http://forum.xda-developers.com/showthread.php?t=2329228](http://forum.xda-developers.com/showthread.php?t=2329228)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,16 +7,16 @@
 ## Getting started
 **1. Add the build system, and the wanted sources to your manifest.**
 
-Find your manifest file. Check `${ANDROID_BUILD_TOP}/.repo/manifest/`
-
-and add the following towards the end:
+Create the file `${ANDROID_BUILD_TOP}/.repo/local_manifests/opengapps.xml` (you might have to create the `local_manifests` folder first) with the following content:
 ```xml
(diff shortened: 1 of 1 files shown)


Blystad:
Hey, 

Thanks for the contribution.

How will this be if a team is working together on a ROM? Does each team member need to do this, or is it checked into git?

fonic:
In a project I was involved we put the overlay manifest on GitHub and initialized the repo like this (nice and easy):

```
# Initialize repository
repo init -u https://android.googlesource.com/platform/manifest -b master

# Add local manifest
mkdir -p .repo/local_manifests
curl https://<github>/overlay.xml > .repo/local_manifests/overlay.xml

# Sync repo
repo sync
```

Blystad:
I think I'm a bigger fan of both alternatives being presented to the user. 

Maybe your local_manifest solution first, and then the custom manifest.xml solution afterwards. Thoughts?

fonic:
I think that's a great idea, then the user can decide which solution fits his needs.

Blystad:
Closing this since it hasn't been updated. Feel free to re-open in the future with the necessary changes :)

PR #31, opened 2016-03-20
Minor code refactoring

### LINKED ISSUE #29: Check for PRODUCT_AAPT_* value in device.mk required
I think opengapps should check if `PRODUCT_AAPT_*` values are present/set in device.mk.

I integrated opengapps in my AOSP build for the HiKey reference board (device/linaro/hikey) and encountered strange build errors because the device.mk for that device lacks those values. If they are missing, `vendor/google/build/core/find_apk.sh` receives wrong arguments and fails to locate packages.

Dudulle:
Quote all shell variables, except when required not to (eg: after find -L)
Trim spaces in makefile function arguments to maintain compatibility with shell commands while maintaining a consistent code style.
Use consistent indentation.

resolves #29

### DIFF
--- a/core/definitions.mk
+++ b/core/definitions.mk
@@ -1,13 +1,13 @@
 define get-allowed-api-levels
-$(shell seq 1 $(PLATFORM_SDK_VERSION))
+$(shell seq 1 "$(PLATFORM_SDK_VERSION)")
 endef
 
 define get-allowed-api-levels-regex
-($(call subst, ,|,$(call get-allowed-api-levels)))
+($(call subst,,|,$(call get-allowed-api-levels)))
 endef
 
 define find-apk-for-pkg
-$(shell $(GAPPS_BUILD_SYSTEM_PATH)/find_apk.sh "$(GAPPS_SOURCES_PATH)" $(PLATFORM_SDK_VERSION) $(PRODUCT_AAPT_PREF_CONFIG) $(2) $(1) $(GAPPS_FORCE_MATCHING_DPI) $(GAPPS_AAPT_PATH))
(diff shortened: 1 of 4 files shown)


core/definitions.mk
```diff
@@ -7,7 +7,7 @@ define get-allowed-api-levels-regex
 endef
 
 define find-apk-for-pkg
-$(shell $(GAPPS_BUILD_SYSTEM_PATH)/find_apk.sh "$(GAPPS_SOURCES_PATH)" $(PLATFORM_SDK_VERSION) $(PRODUCT_AAPT_PREF_CONFIG) $(2) $(1) $(GAPPS_FORCE_MATCHING_DPI) $(GAPPS_AAPT_PATH))
+$(shell $(GAPPS_BUILD_SYSTEM_PATH)/find_apk.sh "$(GAPPS_SOURCES_PATH)" $(PLATFORM_SDK_VERSION) "$(PRODUCT_AAPT_PREF_CONFIG)" $(2) $(1) $(GAPPS_FORCE_MATCHING_DPI) $(GAPPS_AAPT_PATH))
```
mfonville: Probably it would be smart to quote all arguments :-)

Dudulle: Only the last argument can still be empty and that won't cause any issue.
If ever another argument is empty, it would be because there is an error somewhere else.

mfonville: True, I know. But it is a good habit to quote (maintaining all the regular Open GApps buildscripts that use posix shell, I know how much trouble can be avoided by using them consistently ;-) )

fonic:
I vote for quotes for all variables as this might also fix #32.

core/definitions.mk
fonic: I vote for quotes for all variables as this might also fix #32.

Blystad: +1 on quotes for all variables.

Dudulle:
Updated.

If we're going to quote everything, at least let's really quote everything :)

PR #33, opened 2016-03-24
Avoid too many open files error

Dudulle:
Including CLEAR_VARS inside GAPPS_CLEAR_VARS seems to cause issues on some builds.
This separates the two files to solve the issue.

### DIFF
--- a/core/clear_vars.mk
+++ b/core/clear_vars.mk
@@ -1,5 +1,3 @@
-include $(CLEAR_VARS)
-
 # Reset custom local variables.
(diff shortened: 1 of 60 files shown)


sevenrock:
The problem I reported here (https://github.com/opengapps/aosp_build/pull/28#issuecomment-200550883) is fixed by this PR.

niks255:
Yeah it did fix my problem too..

Blystad:
I'll merge this for now, but we should be able to solve this with a Makefile macro.

Issue #35, opened 2016-03-24
SetupWizard Crashes Immediately


aaboyd:
Any ideas welcome!

Not sure what would cause this,  I have tried with `GAPPS_FORCE_MATCHING_DPI := true` and `GAPPS_FORCE_MATCHING_DPI := false` .

```
E/AndroidRuntime( 1981): FATAL EXCEPTION: main
E/AndroidRuntime( 1981): Process: com.google.android.setupwizard, PID: 1981
E/AndroidRuntime( 1981): java.lang.RuntimeException: Unable to start activity ComponentInfo{com.google.android.setupwizard/com.google.android.setupwizard.WelcomeActivity}: android.content.res.Resources$NotFoundException: Resource ID #0x1110038
E/AndroidRuntime( 1981): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2358)
E/AndroidRuntime( 1981): 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2420)
E/AndroidRuntime( 1981): 	at android.app.ActivityThread.access$900(ActivityThread.java:154)
E/AndroidRuntime( 1981): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1321)
E/AndroidRuntime( 1981): 	at android.os.Handler.dispatchMessage(Handler.java:102)
E/AndroidRuntime( 1981): 	at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime( 1981): 	at android.app.ActivityThread.main(ActivityThread.java:5294)
E/AndroidRuntime( 1981): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime( 1981): 	at java.lang.reflect.Method.invoke(Method.java:372)
E/AndroidRuntime( 1981): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
E/AndroidRuntime( 1981): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
E/AndroidRuntime( 1981): Caused by: android.content.res.Resources$NotFoundException: Resource ID #0x1110038
E/AndroidRuntime( 1981): 	at android.content.res.Resources.getValue(Resources.java:1343)
E/AndroidRuntime( 1981): 	at android.content.res.Resources.getValue(Resources.java:1323)
[...]

PR #38, opened 2016-04-06
Configure network location provider overlay for all types

### LINKED ISSUE #12: Need to patch DefaultPermissionGrantPolicy.java for GApps on 6.0
If we build gapps in aosp 6.0 source tree, a patch should be applied as  [Notes-for-Android-6.0](https://github.com/opengapps/opengapps/wiki/Notes-for-Android-6.0) saying

```
cd frameworks/base
curl https://github.com/TeamExodus/frameworks_base/commit/9c36be651e83fb039a262682839bd920b033007a.patch | patch -p1
```

Blystad:
The network location provider must be configured in the XML for all
device types, otherwise issues occur on 6.0+ with permissions, and
on all devices with the network location provider.

I'm not sure how this will go in CyanogenMod builds since there
vendor/cm already defines a similar overlay...It should go OK thought.

This should fix #12 and #35

### DIFF
--- a/opengapps-packages.mk
+++ b/opengapps-packages.mk
@@ -1,6 +1,8 @@
 include vendor/google/build/config.mk
 include $(GAPPS_FILES)
 
+DEVICE_PACKAGE_OVERLAYS += $(GAPPS_DEVICE_FILES_PATH)/overlay/pico
+
 PRODUCT_PACKAGES += GoogleBackupTransport \
                     GoogleContactsSyncAdapter \
                     GoogleFeedback \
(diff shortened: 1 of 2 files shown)


Dudulle:
When the overlays contains the same attributes, the last overlay will override the previous value of that attribute. So there should not be any problem with CyanogenMod.

But this will probably not fix #35 since, as mentioned there, the error seems to be the setup package name. (I have the location provider set and got the same error last week after doing a clean install.

Blystad:
Blergh, you're right, it probably will just fix #12 . I'll see if I can manage to make a PR to fix #35 tomorrow.

PR #40, opened 2016-04-07
Fix SetupWizard package name is updated based on tablet/phone

Blystad:
This fixes issue #35

### DIFF
--- a/modules/SetupWizard/Android.mk
+++ b/modules/SetupWizard/Android.mk
@@ -2,7 +2,12 @@ LOCAL_PATH := .
 include $(CLEAR_VARS)
 include $(GAPPS_CLEAR_VARS)
 LOCAL_MODULE := SetupWizard
-LOCAL_PACKAGE_NAME := com.google.android.setupwizard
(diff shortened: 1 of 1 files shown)


modules/SetupWizard/Android.mk
```diff
@@ -2,7 +2,12 @@ LOCAL_PATH := .
 include $(CLEAR_VARS)
 include $(GAPPS_CLEAR_VARS)
 LOCAL_MODULE := SetupWizard
-LOCAL_PACKAGE_NAME := com.google.android.setupwizard
+ifneq ($(filter $(PRODUCT_CHARACTERISTICS),tablet),)
+	LOCAL_PACKAGE_NAME := com.google.android.setupwizard.tablet
+else
+	LOCAL_PACKAGE_NAME := com.google.android.setupwizard.phone
```
mfonville: wrong!
should be .default

Blystad: Just checking that you were awake ;) (Let's at least go with that explanation...). Anyways, fixed.

Gianmariai: :)

Issue #37, opened 2016-04-06
Make targets fail due to GAPPS_VARIANT not being set


fonic:
When OpenGApps is part of the source tree, several make targets fail, e.g. `make clean`. This is due to the fact that BoardConfig.mk or device.mk is not included when making these targets, thus GAPPS_VARIANT is undefined.

I would suggest the the following patch for `vendor/google/build/config.mk`:

```
# Device should define their GAPPS_VARIANT in device/manufacturer/product/BoardConfig.mk
ifeq ($(GAPPS_VARIANT),)
  #$(error GAPPS_VARIANT must be configured)
  # If GAPPS_VARIANT is not set, default to variant 'stock'
  GAPPS_VARIANT := stock
endif
GAPPS_VARIANT_EVAL := $(call get-gapps-variant,$(GAPPS_VARIANT))
```

Also, it should probably print a note stating that GAPPS_VARIANT was not set and now defaults to 'stock'. Since 'stock' is the recommended variant anyway, I think this something people can live with.

Blystad:
I had this originally, when `GAPPS_VARIANT` was set in the `BoardConfig.mk` file, but that presented an issue that made me move `GAPPS_VARIANT` to `device.mk` and require it to be set (mainly that if `GAPPS_VARIANT` was set to e.g. `mini` in `BoardConfig.mk` then it wouldn't be set in `device.mk`, causing the `include` of `opengapps-packages` to provide all packages.

But I can see the issue....I don't think defaulting to stock is a good solution though. I suggest rather that we default to `none` and print a warning.

`none` doesn't exist yet, but should be fairly straightforward to add.

fonic:
Defaulting to 'none' also sounds like a good idea.

jamuir:
@Blystad:  do you recall the reasoning as to why the file `vendor/google/build/config.mk` is used?

That file is treated specially by the build system.  It is included by `build/core/main.mk`:

```make
# Include the google-specific config
-include vendor/google/build/config.mk
```

It also gets included by the file `vendor/google/build/opengapps-packages.mk`.

So `config.mk` is processed at least twice when a build is done.

I think the issue with `make installclean` and `make clean` could be avoided if we rename `config.mk` to something else.

Blystad:
Originally it was to ensure that `mm` in `vendor/google/build/modules/*/` worked. I don't know if it's still needed.

jamuir:
ok.  I will do some testing.  There may be another way to get `mm` working.

[...]

PR #41, opened 2016-04-10
Reorganize files to allow placing the repository anywhere.

Dudulle:
OpenGApps source can be specified using GAPPS_SOURCES_PATH (default value: vendor/opengapps/sources)

Both opengapps-packages.mk and the aosp build system were including config.mk.
Including it from the build system is not necessary.
By not relying on the aosp build system to include it, we can move the repository and add the google apps only on specific devices/configurations.

Note: after this, device.mk should include opengapps.mk instead of opengapps-packages.mk

Since the "config.mk" trick is not used anymore, this should also solve #37 (been using this for a while now without any issue)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ### Disclaimer
 1. *Use this at your own risk.* Cyanogenmod received a cease and desist letter from Google when they included Google Apps in their ROM. See: [A Note on Google Apps for Android](http://android-developers.blogspot.com/2009/09/note-on-google-apps-for-android.html)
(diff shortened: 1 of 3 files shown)


Blystad:
This breaks building a single module with `mm`. The definitions won't be included. For that we use the hook in the AOSP build system.

Dudulle:
As long as the build is setup for a device where opengapps is included, it is included and it is still possible to build a single module with either make or mm.

Maybe this was an issue before and has been fixed by a previous commit?

Blystad:
Hey,

Sorry that I haven't gotten the time to properly test this. But, I think now that we've started to use overlays (and there will be many more overlays in the future), I'm going to close this PR.

If we put overlays in a non-whitelisted vendor directory, this should fail:
https://android.googlesource.com/platform/build/+/master/core/tasks/vendor_module_check.mk#80

PR #50, opened 2016-05-25
build: Do not force to use stock if user wants WebViewGoogle

alviteri:
in the case if the user is using another gapps module different than stock and wants use WebViewGoogle this will be affected
by missing overlay for it .... move it to proper .mk for make it call every time when WebViewGoogle is set in any gapps module

### DIFF
--- a/modules/WebViewGoogle/Android.mk
+++ b/modules/WebViewGoogle/Android.mk
@@ -6,6 +6,8 @@ LOCAL_PACKAGE_NAME := com.google.android.webview
 
 LOCAL_OVERRIDES_PACKAGES := webview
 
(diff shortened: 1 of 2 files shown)


Blystad:
Heya,

This won't work. You're not allowed to modify DEVICE_PACKAGE_OVERLAYS in a module, (technically, not any PRODUCT_ or DEVICE_ variables. Check `assert-product-vars` in AOSP build system, at least on LP) and also this would be called always (since all .mk files are scanned), so it would always be assigned (if it was possible, which it is not :) ).

alviteri:
thank you .... I was confused due to https://github.com/opengapps/aosp_build/blob/master/opengapps-packages.mk#L64 anyway thanks man ...... then I can close this

PR #57, opened 2016-05-31
Allow an easy way to include WebViewGoogle on non-stock builds

AdFad666:
We need to include an overlay for frameworks so that it can use
the Google built webview. Use GAPPS_FORCE_WEBVIEW_OVERRIDES in
your device.mk file.


Blystad:
Thanks!

PR #61, opened 2016-06-16
Chrome: also override Browser2

AdFad666:
https://android.googlesource.com/platform/packages/apps/Browser2/

This is used on AOSP Master branches

### DIFF
--- a/modules/Chrome/Android.mk
+++ b/modules/Chrome/Android.mk
@@ -5,6 +5,6 @@ LOCAL_MODULE := Chrome
 LOCAL_PACKAGE_NAME := com.android.chrome
 
 GAPPS_LOCAL_OVERRIDES_MIN_VARIANT := stock
-GAPPS_LOCAL_OVERRIDES_PACKAGES := Browser BrowserProviderProxy Chromium Fluxion Gello
+GAPPS_LOCAL_OVERRIDES_PACKAGES := Browser Browser2 BrowserProviderProxy Chromium Fluxion Gello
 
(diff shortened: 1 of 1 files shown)


Blystad:
Thanks. Nice to see Google finally substituted the old AOSP browser with something a bit more useful.

AdFad666:
I wouldn't call it useful! :joy_cat: 

device-2016-06-16-214725

Blystad:
Wow...Forgive me for having hope... :)

alviteri:
@AdFad666 you were more fast than me lol @Blystad is this project able to be builded in L too? .... thanks

Blystad:
@alviteri I tested quickly and nope, there's some permission checks as well as SDK constants pointing to M or newer. But, if you delete them then probably.

Though I guess the better alternative is to just use a prebuilt Chromium APK if you want an open source browser. AFAIK it should be pretty similar to Chrome now.

alviteri:
@Blystad thanks ... then I will continue using gapps zip for L xD

Blystad:
Well, aosp_build will work for L, it's just this particular app (Browser2) won't :) I use aosp_build with L every day.

alviteri:
@Blystad Calculator and Google Dialer cause errors in build like

mv: cannot stat ‘/mnt/users/alviteri/aospl/out/target/product/honami/obj/APPS/GoogleDialer_intermediates/package.apk’: No such file or directory

they need another special think or they are not allowed to be builded in L .... because M/N there are not problems

PR #62, opened 2016-06-17
Prevent GoogleCalculator be called by L

alviteri:
…led by L


alviteri:
GoogleTTS is already called  by pico https://github.com/opengapps/aosp_build/blob/master/opengapps-packages.mk#L19

also if you set GoogleCalculator package in your L enviroment then it will be called and the build will crash

Blystad:
GoogleTTS isn't doubled. In the base package it's included for M and newer (since the setup wizard needs it), while the other check is if its L or older.

So, please revert the GoogleTTS part.

Calculator Google, sure, I would love to merge that part. :)

alviteri:
ok

PR #64, opened 2016-06-18
Allow override AOSP Dialer

alviteri:
Using this system you will use GoogleDialer as stock and override AOSP Dialer .... the patches practically
let you to use InCallUI without have AOSP Dialer or it is impossilble .....

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,12 @@ If you want to include WebViewGoogle on a non-stock build you need:
 GAPPS_FORCE_WEBVIEW_OVERRIDES := true
 ```
 
+If you want to include Google Dialer on a non-stock build you need:
+
+```
+GAPPS_FORCE_DIALER_OVERRIDES := true
+```
+
 On a per-app basis, add the GApps package to `GAPPS_PACKAGE_OVERRIDES`.
 Example:
 
--- a/opengapps-packages.mk
+++ b/opengapps-packages.mk
@@ -68,6 +68,9 @@ PRODUCT_PACKAGES += Books \
 ifneq ($(filter $(TARGET_GAPPS_VARIANT),stock),) # require at least stock
 
 GAPPS_FORCE_WEBVIEW_OVERRIDES := true
+ifneq ($(filter $(call get-allowed-api-levels),23),)
+GAPPS_FORCE_DIALER_OVERRIDES := true
+endif
 
 PRODUCT_PACKAGES += GoogleCamera \
                     GoogleContacts \
@@ -104,4 +107,11 @@ endif # end nano
 ifeq ($(GAPPS_FORCE_WEBVIEW_OVERRIDES),true)
 DEVICE_PACKAGE_OVERLAYS += $(GAPPS_DEVICE_FILES_PATH)/overlay/webview
 PRODUCT_PACKAGES += WebViewGoogle
+endif
+
+ifneq ($(filter $(call get-allowed-api-levels),23),)
+ifeq ($(GAPPS_FORCE_DIALER_OVERRIDES),true)
+DEVICE_PACKAGE_OVERLAYS += $(GAPPS_DEVICE_FILES_PATH)/overlay/dialer
+PRODUCT_PACKAGES += GoogleDialer
+endif
 endif
--- a/overlay/dialer/packages/services/Telecomm/res/values/config.xml
+++ b/overlay/dialer/packages/services/Telecomm/res/values/config.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (C) 2014 The Android Open Source Project
+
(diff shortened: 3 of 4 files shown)


alviteri:
@Blystad

Blystad:
I don't truly see the need of the two patches...

```
-            return new ComponentName(
 +-                    resources.getString(R.string.ui_default_package),
 +-                    resources.getString(R.string.dialer_default_class))
```

Isn't it enough to just override `ui_default_package` and `dialer_default_class` from AOSPs values to Google values?

This patch is a necessity IF you want to use EITHER AOSP Dialer or Google Dialer, depending on what's installed. In the aosp_build case, only one of them will be installed.

alviteri:
Yeah but the support will be for both in case user by A or B prefer use aosp one

alviteri:
Anyway I am using this on my aosp builds using this gapps and there are not problems

Blystad:
Thing is, I would like to keep patches to an absolute minimum. They should be the last resort. Normal stock AOSP builds should work fine with Gapps without any patches.

In this case, I see the patches more as general convinience to the ROM community, rather than a necessity for making the dialer work. When building the GoogleDialer you end up with [either the GoogleDialer or the AOSPDialer](https://github.com/opengapps/aosp_build/blob/master/modules/GoogleDialer/Android.mk) because of the override. Never both.

I would welcome a patch that uses the existing overlay infrastructure to assign the proper values to `ui_default_package` and `dialer_default_class`.

mfonville:
Otherwise they could also be put in e.g. the wiki

alviteri:
Arne in that case can you show me how do you do...  Because building Googledialer I can't see the InCallUI also giving all the permissions....  Non without those patches

Blystad:
@alviteri sure. See this for inspiration: https://github.com/opengapps/aosp_build/blob/master/overlay/webview/frameworks/base/core/res/res/values/config.xml

Create a new folder in `overlay/dialer`

Notice the file path: `frameworks/base/core/res/res/values` in the above example. That must be `packages/services/Telecomm/res/values` and `packages/services/Telephony/res/values` for you. (so, you'll need two `config.xml` files).

Then inside these folders you need to create a `config.xml` file. You'll want to override these two config keys:
```
    <!-- Package name for the default in-call UI and dialer [DO NOT TRANSLATE] -->
    <string name="ui_default_package" translatable="false">com.android.dialer</string>

    <!-- Class name for the default main Dialer activity [DO NOT TRANSLATE] -->
    <string name="dialer_default_class" translatable="false">com.android.dialer.DialtactsActivity</string>
```

In both cases to point to the Google values. (See your current patches, they have them :) ). Then you'll want to include the `overlay/dialer` in the stock+ configuration (see how WebView does it), and in your FORCE configuration.

alviteri:
Are you sure..  In that way InCallUI will be show while calling. .  If yes thanks I didn't know then the patches are not needed...  I Will update the PR when in home

Blystad:
In Telecomm you need to override this in addition to the two above:

```

    <!-- Class name for the default in-call UI Service [DO NOT TRANSLATE] -->
    <string name="incall_default_class" translatable="false">com.android.incallui.InCallServiceImpl</string>

```

Then I think we're there.

alviteri:
@Blystad but the changes in classes are still needed?  Confused 😕 xD

Blystad:
You just need to create two overlay XML files :) No patches necessary.

alviteri:
ok check now

alviteri:
build and ready +1

Blystad:
You need to override a few more config values :)

In both config.xml you need to add `ui_default_package` and assign it to `com.google.android.dialer`.

Based on the existing values, I think `dialer_default_class` should be just `com.google.android.dialer.extensions.GoogleDialtactsActivity` and likewise `incall_default_class` should be just `com.android.incallui.InCallServiceImpl`.

It might work fine with the current values, but I would prefer to be as correct as possible. With those changes, I'll merge it.

alviteri:
updated

Blystad:
On further review:

```
+    <!-- Class name for the default in-call UI Service [DO NOT TRANSLATE] -->
 +    <string name="incall_default_class" translatable="false">com.android.incallui.InCallServiceImpl</string>
```

Doesn't this need to point to something in the com.google.android.dilaer package?

alviteri:
but you said

likewise incall_default_class should be just com.android.incallui.InCallServiceImpl

Blystad:
Yes, but I didn't see at first that it just pointed to the `com.android.incallui.InCallService` at first - minor mistake on my end.

But, I see now that the Google guys just has a service in the APK with the exact same name. So it should be fine.

Please delete `incall_default_class` as it's unnecessary. Then I'll merge it. Sorry for the confusion :)

alviteri:
screenshot_20160621-100651

yeah seems not needed .... PR updated

Blystad:
Thanks for your contribution. Merged :)

Issue #63, opened 2016-06-18
Build problem with Nexus 5X (possibly Nexus 6P too) due to new vendor partition


CaseyBakey:
Hello there,

I did follow the instructions on the readme to integrate opengapps into my AOSP build but I'm encountering problems:

`Target system fs image: out/target/product/bullhead/obj/PACKAGING/systemimage_intermediates/system.img
Non-symlink out/target/product/bullhead/system/vendor detected!
You cannot install files to out/target/product/bullhead/system/vendor while building a separate vendor.img!
make: *** [out/target/product/bullhead/obj/PACKAGING/systemimage_intermediates/system.img] Error 1
make: *** Waiting for unfinished jobs....
`

I guess it has to do with the new partition layout of N5X/N6P:

http://forum.xda-developers.com/nexus-6p/general/developer-discussion-flashing-vendor-t3246767

I did try to edit every scripts from this repo mentioning /system/vendor, to point to /vendor instead, but it wasn't enough.

I'm not a git guru nor an Android one (first AOSP build attempt here), so I'll be glad to receive some help here ;)

Thanks again for your work!

mfonville:
@Blystad for devices that have their own `/vendor` partition we should probably not touch it (and users should not flash it, only flashing their `/system`)

CaseyBakey:
So what should I do exactly?

Blystad:
Try to delete `vendor/pittpatt `in `opengapps-files.mk` and see if it works then.

If it does we just need to find which board variables is set, and exclude this folder in cases where the device has a vendor partition

CaseyBakey:
Ok, I have first to finish a build and then I'll try.
But deleting pittpatt will break facial recognition, right? I'm not using it but just to know.

Blystad:
Your vendor.img will probably already contain them. So as long as you don't
flash that, you should be OK.

CaseyBakey:
Ok, perfect. Thanks for your fast answser, I'll keep you updated!

mfonville:
@Blystad @CaseyBakey I can indeed (from the regular Open GApps packages) confirm that if you have flashed your `/vendor` from the regular stock images (and you should update it sometimes) the files should already be there.

Having `/vendor` separate is imho btw a good new practice set by Google that should probably become the default for other devices too.

Blystad:
[...]

PR #65, opened 2016-06-20
Don't copy vendor/pittpatt on device with separate vendor.img

Blystad:
This should fix build issues with Nexus 5X

Refs: #63


Blystad:
Closing this since it didn't seem to fix the issue. Also seems like vendor/pittpatt disappears on Android N.

PR #66, opened 2016-06-21
Allow override AOSP Browser

alviteri:
also include needed overlay for frameworks

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,12 @@ If you want to include Google Dialer on a non-stock build you need:
 GAPPS_FORCE_DIALER_OVERRIDES := true
 ```
 
+If you want to include Chrome on a non-full build you need:
+
+```
+GAPPS_FORCE_BROWSER_OVERRIDES := true
+```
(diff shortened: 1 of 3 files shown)


alviteri:
@Blystad this still will let you change default browser in settings-apps

Blystad:
Seems I was a bit too fast in merging this in:

```
vendor/google/build/overlay/browser/frameworks/base/core/res/res/values/config.xml:29: error: Resource at default_browser appears in overlay but not in the base package; use <add-resource> to add.
```

Blystad:
I guess the overlay should only be included on Marshmallow and above. On Lollipop the key doesn't exist.

alviteri:
only .... that ?.... sorry I didn't test in L

Blystad:
I pushed 1acbc75c613ba6ceb19f0b383fe7150bcf885123 to fix it :)

alviteri:
thanks .... and sorry for not check L

PR #68, opened 2016-06-21
find_apk: Fix compatibility with Bash 3.2

Blystad:
Lollipop is, by recommendation, built on Ubuntu 12.04 where Bash 3.2
is the latest.

declare -n was added in Bash 3.3.

CC: @Dudulle - since you made the change I would love if you could take a look. I tested it on Lollipop and it works fine. I don't see how `currapk` was actually needed.

### DIFF
--- a/core/find_apk.sh
+++ b/core/find_apk.sh
@@ -10,7 +10,6 @@ getlatestapk() {
   sourceapk=""
   highcompatversion=""
   tempapks=( )
-  declare -n currapk="tempapks$i"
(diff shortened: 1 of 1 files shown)


Dudulle:
currapk was only a pointer to the last element of the list.
As you said, it wasn't really needed in this case.
This change should have the exact same effect.

Dudulle:
I ran a few tests with bash 4.3 and I did not notice any difference between the two methods.
This should be safe to merge.

PR #69, opened 2016-07-26
Always use the highest compatible API level.

Dudulle:
Some apps (at least Play Services) will not work if they have multiple api levels and the one installed is not the highest API the device can take.
This cannot be detected from aapt because the min and target sdk in the manifest are the same.
It is therefore better to favor the correct API level instead a more recent package which might not work.

### DIFF
--- a/core/find_apk.sh
+++ b/core/find_apk.sh
@@ -2,56 +2,47 @@
 
 getlatestapk() {
   # This functions finds the latest available version for the given architeture.
-  if ! stat --printf='' "$SOURCES/$2/"*"app/$1" 2> /dev/null; then
-    return 1 #appname is not there, error!?
-  fi
-  appname="$1"
-  aaptcmd="$4"
+  aaptcmd="$3"
   sourceapk=""
-  highcompatversion=""
+  highversion=""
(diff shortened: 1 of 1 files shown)


Blystad:
I'm on vacation for a couple of weeks, but don't we already handle this? The opengapps project places apps into folders based on API level, and we take the one most fitting for our API level.

Dudulle:
For the moment, the code will skip API's that are higher than the device's API, but not the ones that are lower in some circumstances.

The code will select the latest version regardless of the API level. If the highest API the device supports has not been updated or a lower API has a better (updated) DPI for the device, the lower API level will be selected.

I got this multiple times with Play Services when building for Marshmallow and 240dpi.
Last time I had this problem, those apk's had the latest version: API23-480dpi (the one I should have installed), API21-320dpi, API9-320&240dpi.
The order of priority for my device was then: API9-240dpi, API21-320dpi, API9-320dpi, API23-480dpi.
With this commit and in the same situation, the API23-480dpi would be the only one available (Except when forcing the dpi, it would then take the highest version between 240dpi and nodpi, but still one for API 23)

PR #74, opened 2016-08-25
remove webview overlay from N builds

alviteri:
it is not defined in frameworks in Android N

Signed-off-by: David Viteri <davidteri91@gmail.com>

PR #75, opened 2016-08-31
webview: add webview overlay for N

alviteri:

### DIFF
--- a/opengapps-packages.mk
+++ b/opengapps-packages.mk
@@ -107,7 +107,9 @@ endif # end micro
 endif # end nano
 
 ifeq ($(GAPPS_FORCE_WEBVIEW_OVERRIDES),true)
-ifneq ($(filter-out $(call get-allowed-api-levels),24),)
+ifneq ($(filter $(call get-allowed-api-levels),24),)
+DEVICE_PACKAGE_OVERLAYS += $(GAPPS_DEVICE_FILES_PATH)/overlay/webview-24
+else
 DEVICE_PACKAGE_OVERLAYS += $(GAPPS_DEVICE_FILES_PATH)/overlay/webview
 endif
 PRODUCT_PACKAGES += WebViewGoogle
--- a/overlay/webview-24/frameworks/base/core/res/res/xml/config_webview_packages.xml
+++ b/overlay/webview-24/frameworks/base/core/res/res/xml/config_webview_packages.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+/*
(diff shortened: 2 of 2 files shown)


mfonville:
You might want to check out this comment, we have more information from the Google Stock ROM to add Chrome (and its variants) as Webview providers too: https://github.com/opengapps/aosp_build/pull/74#issuecomment-243563980

alviteri:
then I need use it ?

mfonville:
On Android Nougat Chrome will be (if installed & enabled, at least on Stock ROMs) be used as the primary webview provider. So probably it would be best to achieve feature-parity with that.
@TheCrazyLex is cooking up a patch for it, so you might want to wait for it and then combine it for the overlay.

alviteri:
my build is at 99% let me see how this actual patch works

opengapps-packages.mk
```diff
@@ -107,7 +107,9 @@ endif # end micro
 endif # end nano
 
 ifeq ($(GAPPS_FORCE_WEBVIEW_OVERRIDES),true)
-ifneq ($(filter-out $(call get-allowed-api-levels),24),)
+ifneq ($(filter $(call get-allowed-api-levels),24),)
+DEVICE_PACKAGE_OVERLAYS += $(GAPPS_DEVICE_FILES_PATH)/overlay/webview-n
```
AdFad666: I suggest `webview-24` to match the API level, everything else LGTM

alviteri:
@mfonville 
screenshot_20160831-150533

alviteri:
for @mfonville @Blystad 

those patches can be temporal fix for built-in opengapps with Nougat

https://github.com/SonyAosp/platform_frameworks_base/commit/3bd6ba0dee6f32e4e52a4bb58474ba4f4f71edd0

https://github.com/SonyAosp/platform_frameworks_base/commit/97da3fac0d6d92d22bec5b83de76404b53f8a765

https://github.com/SonyAosp/platform_frameworks_base/commit/5a6d9c0a4e2aea743a3836bc5b3fdf8d5d54be3b

alviteri:
updated patch as @AdFad666 request

alviteri:
will test https://github.com/opengapps/aosp_build/pull/74#issuecomment-243563980 and see

Blystad:
The framework patches should be unnecessary with the overlays that we use. The overlays should tell the package manager that this package should be granted these permissions. That's what Google does on Nexus devices. I've decompiled the code to check it.

Also, I've fixed this and more in PR #76. Please take a look.

alviteri:
ok then closing this

PR #76, opened 2016-08-31
Fix nougat

### LINKED ISSUE #73: Changes for Nougat
The following changes may need to be ported to aosp_build:
* New config file `/etc/sysconfig/google_vr_build.xml` (all Android versions)
* `/vendor/pittpatt/` seems to be dropped in Nougat
* `libfacelock_jni.so` has been renamed to `libfacenet.so` (Nougat only)
* the keyboard swypelibs have been dropped (Nougat only)
* New app `com.google.vr.vrcore` (all Android versions)
* Path of Google Keyboard has changed to `LatinIMEGooglePrebuilt` (Nougat only)
* New apps `com.google.android.printservice.recommendation`, `com.google.android.ext.shared` and `com.google.android.ext.services` (Nougat only)

Blystad:
With this, Google Play now works on Nougat. I've tested this on an arm based device, with the `pico` GAPPS_VARIANT.

I haven't moved LatinIME to the new path...it really shouldn't matter...the path should just be used for directory scanning in the package manager, and nothing more. If it turns out to be needed then I'll figure something out.

Fixes #73 
Supersede #75


TheCrazyLex:
The issue on the webview thing here is that not all packages include chrome stable. Also chrome debug should be removed for security reasons.

modules/SetupWizard/Android.mk
```diff
@@ -9,5 +9,6 @@ else
 endif
 
 LOCAL_PRIVILEGED_MODULE := true
+LOCAL_OVERRIDES_PACKAGES := Provision
```
mfonville: On Marshmallow there was no conflict?

Blystad: The Provision app is just a minimal setup wizard. It just sets a few
properties that the Google SetupWizard also sets. So it's safe to remove on
all versions of Android.

See:
https://android.googlesource.com/platform/packages/apps/Provision/+/356bff898c6f33334d20ea68bd5c49f7a2ef9b6b/src/com/android/provision/DefaultActivity.java

mfonville: ok thanks.
I earlier had disabled its removal on Marshmallow, as it was suggested as a workaround for weird CMSetupwizard behaviour

alviteri:
after build with this https://github.com/alviteri/aosp_build/commits/master there is not webview and play services FC at start without those base patch

Blystad:
Give me logcat output and I'll look on friday.

It works great for me in Pico config, with SetupWizard, without any hacky
patches. patches. The hacky patches fixes the symptome, not the problem.

alviteri:
I am using micro

alviteri:
this is my config

https://github.com/SonyAosp/vendor_google/blob/n/products/gms.mk#L42

Blystad:
Still, I need logcat. Or even better - output of "adb bugreport". With the
bugreport i can see all permissions granted.

alviteri:
https://drive.google.com/file/d/0B-e5eQvKq2t2NkgtZW1peVlkdGc/view

Blystad:
Thanks. I'll check on friday. Got a funeral tomorrow so I'm busy.

alviteri:
Ok tested many many times with micro Gapps Packages and every time after wipe data the webview isn't included at all ..........

alviteri:
ok more news.... I just drope WebView Flag from build in AOSP and Gapps Works now without FC ...

rapperskull:
I just checked the AOSP source code, and found out that an app can be selected as a WebView provider only if the signature matches the one in _config_webview_packages.xml_ OR the package is installed in /system. This is obviously for security reasons. If someone installs Chrome or Google WebView from the Play Store without it being present in /system, it still can't be selected, because Google does not store the signatures for these apps, since they are pre-installed on Nexus devices. We should update the patch to also include the signatures of these two apps. @mfonville is calculating them and will post them here shortly.

mfonville:
I got the signatures for Chrome and for Webview that are used on Nougat, take care that **ARM and ARM64 use different signatures!**!!

base64 signature of arm/app/com.android.chrome/24/nodpi/275609802.apk:
```
MIIGMAYJKoZIhvcNAQcCoIIGITCCBh0CAQExDzANBglghkgBZQMEAgEFADALBgkqhkiG9w0BBwGg
ggRHMIIEQzCCAyugAwIBAgIJAMLgh0ZkSjCNMA0GCSqGSIb3DQEBBAUAMHQxCzAJBgNVBAYTAlVT
MRMwEQYDVQQIEwpDYWxpZm9ybmlhMRYwFAYDVQQHEw1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKEwtH
b29nbGUgSW5jLjEQMA4GA1UECxMHQW5kcm9pZDEQMA4GA1UEAxMHQW5kcm9pZDAeFw0wODA4MjEy
MzEzMzRaFw0zNjAxMDcyMzEzMzRaMHQxCzAJBgNVBAYTAlVTMRMwEQYDVQQIEwpDYWxpZm9ybmlh
MRYwFAYDVQQHEw1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKEwtHb29nbGUgSW5jLjEQMA4GA1UECxMH
QW5kcm9pZDEQMA4GA1UEAxMHQW5kcm9pZDCCASAwDQYJKoZIhvcNAQEBBQADggENADCCAQgCggEB
AKtWLgDYO6IIrgqWbxJOKdoR8qtW0I9Y4sypEwPpt1TTcvZApxsdyxMJZ2JORland2qSGT2y5b+3
JKkedxiLDmpHpDsz2WCbdxgxRczfey5YZnTJ4VZbH0xqWVW/8lGmPav5xVwnIiJS6HXk+BVKZF+J
cWjAsb/GEuq/eFdpuzSqeYTcfi6idkyugwfYwXFU1+5fZKUaRKYCwkkFQVfcAs1fXA5V+++FGfvj
J/CxURaSxaBvGdGDhfXE28LWuT9ozCl5xw4Yq5OGazvV24mZVSoOO0yZ31j7kYvtwYK6NeADwbSx
DdJEqO4k//0zOHKrUiGYXtqw/A0LFFtqoZKFjnkCAQOjgdkwgdYwHQYDVR0OBBYEFMd9jMIhF1Yl
mn/Tgt9r45jk14alMIGmBgNVHSMEgZ4wgZuAFMd9jMIhF1Ylmn/Tgt9r45jk14aloXikdjB0MQsw
CQYDVQQGEwJVUzETMBEGA1UECBMKQ2FsaWZvcm5pYTEWMBQGA1UEBxMNTW91bnRhaW4gVmlldzEU
MBIGA1UEChMLR29vZ2xlIEluYy4xEDAOBgNVBAsTB0FuZHJvaWQxEDAOBgNVBAMTB0FuZHJvaWSC
CQDC4IdGZEowjTAMBgNVHRMEBTADAQH/MA0GCSqGSIb3DQEBBAUAA4IBAQBt0lLO74UwLDYKqs6T
m8/yzKkEu116FmH4rkaymUIE0P9KaMftGlMexFlaYjzmB2OxZyl6euNXEsQH8gjwyxCUKRJNexBi
GcCEyj6z+a1fuHHvkiaai+KL8W1EyNmgjmyy8AW7P+LLlkR+ho5zEHatRbM/YAnqGcFh5iZBqpkn
Hf1SKMXFh4dd239FJ1jWYfbMDMy3NS5CTMQ2XFI1MvcyUTdZPErjQfTbQe3aDQsQcafEQPD+nqAc
tifKZ0Np0IS9L9kR/wbNvyz6ENwPiTrjV2KRkEjH78ZMcUQXg0L3BYHJ3lc69Vs5Ddf9uUGGMYld
X3WfMBEmh/9iFBDAaTCKMYIBrTCCAakCAQEwgYEwdDELMAkGA1UEBhMCVVMxEzARBgNVBAgTCkNh
bGlmb3JuaWExFjAUBgNVBAcTDU1vdW50YWluIFZpZXcxFDASBgNVBAoTC0dvb2dsZSBJbmMuMRAw
DgYDVQQLEwdBbmRyb2lkMRAwDgYDVQQDEwdBbmRyb2lkAgkAwuCHRmRKMI0wDQYJYIZIAWUDBAIB
BQAwDQYJKoZIhvcNAQEBBQAEggEAm7/k8J2fDfLj8Uc1Gf6xq6cM/Te95S7o2Si3wA1ypP069c7A
ityb3Zq68phJlhpJGZZl+Y8TPF0Gu+d+RbrFckS/kZugJJL6+kJDGXKdm6YwPQECpPHuT2aK7MHM
MxwDy1UhXRJvKM7hSaQYlUpzOi27aY7iYZgoinahiD0zK6UgIo0q1U9IE+G8sBlPZvgdIrelQzY4
Y10S/5oy0Tnc/I5kiNCVECMqhZqLkRBtTTZkTkqzFhNTf7Hrq33tNipwt3gLD/z/IFT4hNy0EpgI
Jj+X2WskeI7q+/TH0pzGzLC7OHeE0gJNBZNaHQpMUEIZCBciomw6KUOgQEHL1vrPyw==
```
base64 signature of arm64/app/com.android.chrome/24/nodpi/275609852.apk:
```
MIIGMAYJKoZIhvcNAQcCoIIGITCCBh0CAQExDzANBglghkgBZQMEAgEFADALBgkqhkiG9w0BBwGg
ggRHMIIEQzCCAyugAwIBAgIJAMLgh0ZkSjCNMA0GCSqGSIb3DQEBBAUAMHQxCzAJBgNVBAYTAlVT
MRMwEQYDVQQIEwpDYWxpZm9ybmlhMRYwFAYDVQQHEw1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKEwtH
b29nbGUgSW5jLjEQMA4GA1UECxMHQW5kcm9pZDEQMA4GA1UEAxMHQW5kcm9pZDAeFw0wODA4MjEy
MzEzMzRaFw0zNjAxMDcyMzEzMzRaMHQxCzAJBgNVBAYTAlVTMRMwEQYDVQQIEwpDYWxpZm9ybmlh
MRYwFAYDVQQHEw1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKEwtHb29nbGUgSW5jLjEQMA4GA1UECxMH
QW5kcm9pZDEQMA4GA1UEAxMHQW5kcm9pZDCCASAwDQYJKoZIhvcNAQEBBQADggENADCCAQgCggEB
AKtWLgDYO6IIrgqWbxJOKdoR8qtW0I9Y4sypEwPpt1TTcvZApxsdyxMJZ2JORland2qSGT2y5b+3
JKkedxiLDmpHpDsz2WCbdxgxRczfey5YZnTJ4VZbH0xqWVW/8lGmPav5xVwnIiJS6HXk+BVKZF+J
cWjAsb/GEuq/eFdpuzSqeYTcfi6idkyugwfYwXFU1+5fZKUaRKYCwkkFQVfcAs1fXA5V+++FGfvj
J/CxURaSxaBvGdGDhfXE28LWuT9ozCl5xw4Yq5OGazvV24mZVSoOO0yZ31j7kYvtwYK6NeADwbSx
DdJEqO4k//0zOHKrUiGYXtqw/A0LFFtqoZKFjnkCAQOjgdkwgdYwHQYDVR0OBBYEFMd9jMIhF1Yl
mn/Tgt9r45jk14alMIGmBgNVHSMEgZ4wgZuAFMd9jMIhF1Ylmn/Tgt9r45jk14aloXikdjB0MQsw
CQYDVQQGEwJVUzETMBEGA1UECBMKQ2FsaWZvcm5pYTEWMBQGA1UEBxMNTW91bnRhaW4gVmlldzEU
MBIGA1UEChMLR29vZ2xlIEluYy4xEDAOBgNVBAsTB0FuZHJvaWQxEDAOBgNVBAMTB0FuZHJvaWSC
CQDC4IdGZEowjTAMBgNVHRMEBTADAQH/MA0GCSqGSIb3DQEBBAUAA4IBAQBt0lLO74UwLDYKqs6T
m8/yzKkEu116FmH4rkaymUIE0P9KaMftGlMexFlaYjzmB2OxZyl6euNXEsQH8gjwyxCUKRJNexBi
GcCEyj6z+a1fuHHvkiaai+KL8W1EyNmgjmyy8AW7P+LLlkR+ho5zEHatRbM/YAnqGcFh5iZBqpkn
Hf1SKMXFh4dd239FJ1jWYfbMDMy3NS5CTMQ2XFI1MvcyUTdZPErjQfTbQe3aDQsQcafEQPD+nqAc
tifKZ0Np0IS9L9kR/wbNvyz6ENwPiTrjV2KRkEjH78ZMcUQXg0L3BYHJ3lc69Vs5Ddf9uUGGMYld
X3WfMBEmh/9iFBDAaTCKMYIBrTCCAakCAQEwgYEwdDELMAkGA1UEBhMCVVMxEzARBgNVBAgTCkNh
bGlmb3JuaWExFjAUBgNVBAcTDU1vdW50YWluIFZpZXcxFDASBgNVBAoTC0dvb2dsZSBJbmMuMRAw
DgYDVQQLEwdBbmRyb2lkMRAwDgYDVQQDEwdBbmRyb2lkAgkAwuCHRmRKMI0wDQYJYIZIAWUDBAIB
BQAwDQYJKoZIhvcNAQEBBQAEggEADz6pz+6uKDrZsBpVwW7+ySIcuEvQw58iL894kTJswJVEuCrz
pQc3pIxtdEZD3hUSy5prFs+zHnW1iuuMIx/vL5OmGsBfnIYJLiqCQ3dlZWqzCLA9lVDl5YO4PjHe
iVtZJG+CYdDuxsg4ZUZVErZKL0wVZv9sTZ7M/m+LA6AQNTyZlbTEleR45+mrZ6worQeHMW0UxTnx
pnlgKnkLm592lv7uigwV+RlGvsx2LkIlVsG3Sr67UEZm2i+n1UvOtpoGN9Ib3FoCC2MWlqgtvxzM
U+P/AOHBa851s1tYFRhbYkrfbJ5zbq7VSLT0RcaPL4NAdXdtwMWysJkcfnpYT3t6qg==
```

base64 signature of arm/app/com.google.android.webview/21/nodpi/275609800.apk:
```
MIIFqAYJKoZIhvcNAQcCoIIFmTCCBZUCAQExDzANBglghkgBZQMEAgEFADALBgkqhkiG9w0BBwGg
ggO/MIIDuzCCAqOgAwIBAgIJANi6DgBQG4ZTMA0GCSqGSIb3DQEBBQUAMHQxCzAJBgNVBAYTAlVT
MRMwEQYDVQQIDApDYWxpZm9ybmlhMRYwFAYDVQQHDA1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKDAtH
b29nbGUgSW5jLjEQMA4GA1UECwwHQW5kcm9pZDEQMA4GA1UEAwwHd2VidmlldzAeFw0xNDA4MDgy
MzIwMjBaFw00MTEyMjQyMzIwMjBaMHQxCzAJBgNVBAYTAlVTMRMwEQYDVQQIDApDYWxpZm9ybmlh
MRYwFAYDVQQHDA1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKDAtHb29nbGUgSW5jLjEQMA4GA1UECwwH
QW5kcm9pZDEQMA4GA1UEAwwHd2VidmlldzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEB
AMbtaFX0r5aZJMAbPVMAgK1ZZ29dTn91VsGxXv2hqrQo7IpqEy2JmPvPnoMsSiuTAe+UcQy8oKDQ
2aYVSAd1DGIy+nSRyFTt3LSIAdwSBkB1qT4a+OqkpsR6bSNXQXQ18lCQu9gREY3h3QlYBQAyzRxw
4hRGlrXAzuSz1Ec4W+6x4nLG5DG61MAMR8ClF9XSqbmGB3kyZ70A0X9OPYYxiMWP1ExaYvpaVqjy
ZZcrPwr+vtW8oCuGBUtHpBUH3OoG+9s2YMcgLG7vCK9awKDqlPcJSpIAAj6uGs4gORmkqxZRMskL
STWbhP4p+3Ap8jYzTVB6Y1/DMVmYTWRMcPW0macCAwEAAaNQME4wHQYDVR0OBBYEFJ6bAR6/QVm4
w9LRSGQiaR5Rhp3TMB8GA1UdIwQYMBaAFJ6bAR6/QVm4w9LRSGQiaR5Rhp3TMAwGA1UdEwQFMAMB
Af8wDQYJKoZIhvcNAQEFBQADggEBAEQu8QiVxax7/diEiJrgKE1LwdXsIygJK/KnaKdnYEkAQpeu
/QmrLiycm+OFbL1qHJIB7OuI/PQBUtcaNSiJSCVgtwtEbZWWIdsynqG/Nf4aGOndXegSQNRH54M0
5sRHLoeRycPrY7xQlEwGikNFR76+5UdwFBQI3Gn22g6puJnVukQm/wXQ+ajoiS4QclrNlixoDQsZ
4STLH4+Wju2wIWKFFArIhVEIlbamq+p6BghuzH3aIz/Fy0YTQKi7SA+0fuNeCaqlSm5pYSt6p5CH
89y1Fr+wFc5r3iLRnUwRcy08ESC7bZJnxV3d/YQ5valTxBbzku/dQbXVj/xg69H8l8MxggGtMIIB
qQIBATCBgTB0MQswCQYDVQQGEwJVUzETMBEGA1UECBMKQ2FsaWZvcm5pYTEWMBQGA1UEBxMNTW91
bnRhaW4gVmlldzEUMBIGA1UEChMLR29vZ2xlIEluYy4xEDAOBgNVBAsTB0FuZHJvaWQxEDAOBgNV
BAMTB3dlYnZpZXcCCQDYug4AUBuGUzANBglghkgBZQMEAgEFADANBgkqhkiG9w0BAQEFAASCAQA4
sdv33hQE0VWxxe/wRqXgjKF3OTS0yO7BkhshxxJn8Nk7KuXzlautZhPssam+oWgL4P3kWtNe6CxA
TbObcwFtv6HENOcg5v7fkXILZYsWEmg7cgmlww2DNS6cCRVlM08HlUJuEbcu6RW6D/dzWLhzu5rK
wi26DpJ9QlNYdGq6xPICa38w1/E+U47LgprRmmUyd1fulHYO/X0uyLuUVc/f6fwDjtjyOTT/yGVE
XLFFbrcj/QMaJs2Zx55ym+iq/eDRrfGlOaYttu/cWMBTUxnNo+fpCrz1SWMqR8i6h2TipI/gXevP
ze0DVZHpvx65nqzBdh/WxM9PKEMCl1YTudir
```
base64 signature of arm64/app/com.google.android.webview/21/nodpi/275609850.apk:
```
MIIFqAYJKoZIhvcNAQcCoIIFmTCCBZUCAQExDzANBglghkgBZQMEAgEFADALBgkqhkiG9w0BBwGg
ggO/MIIDuzCCAqOgAwIBAgIJANi6DgBQG4ZTMA0GCSqGSIb3DQEBBQUAMHQxCzAJBgNVBAYTAlVT
MRMwEQYDVQQIDApDYWxpZm9ybmlhMRYwFAYDVQQHDA1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKDAtH
b29nbGUgSW5jLjEQMA4GA1UECwwHQW5kcm9pZDEQMA4GA1UEAwwHd2VidmlldzAeFw0xNDA4MDgy
MzIwMjBaFw00MTEyMjQyMzIwMjBaMHQxCzAJBgNVBAYTAlVTMRMwEQYDVQQIDApDYWxpZm9ybmlh
MRYwFAYDVQQHDA1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKDAtHb29nbGUgSW5jLjEQMA4GA1UECwwH
QW5kcm9pZDEQMA4GA1UEAwwHd2VidmlldzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEB
AMbtaFX0r5aZJMAbPVMAgK1ZZ29dTn91VsGxXv2hqrQo7IpqEy2JmPvPnoMsSiuTAe+UcQy8oKDQ
2aYVSAd1DGIy+nSRyFTt3LSIAdwSBkB1qT4a+OqkpsR6bSNXQXQ18lCQu9gREY3h3QlYBQAyzRxw
4hRGlrXAzuSz1Ec4W+6x4nLG5DG61MAMR8ClF9XSqbmGB3kyZ70A0X9OPYYxiMWP1ExaYvpaVqjy
ZZcrPwr+vtW8oCuGBUtHpBUH3OoG+9s2YMcgLG7vCK9awKDqlPcJSpIAAj6uGs4gORmkqxZRMskL
STWbhP4p+3Ap8jYzTVB6Y1/DMVmYTWRMcPW0macCAwEAAaNQME4wHQYDVR0OBBYEFJ6bAR6/QVm4
w9LRSGQiaR5Rhp3TMB8GA1UdIwQYMBaAFJ6bAR6/QVm4w9LRSGQiaR5Rhp3TMAwGA1UdEwQFMAMB
Af8wDQYJKoZIhvcNAQEFBQADggEBAEQu8QiVxax7/diEiJrgKE1LwdXsIygJK/KnaKdnYEkAQpeu
/QmrLiycm+OFbL1qHJIB7OuI/PQBUtcaNSiJSCVgtwtEbZWWIdsynqG/Nf4aGOndXegSQNRH54M0
5sRHLoeRycPrY7xQlEwGikNFR76+5UdwFBQI3Gn22g6puJnVukQm/wXQ+ajoiS4QclrNlixoDQsZ
4STLH4+Wju2wIWKFFArIhVEIlbamq+p6BghuzH3aIz/Fy0YTQKi7SA+0fuNeCaqlSm5pYSt6p5CH
89y1Fr+wFc5r3iLRnUwRcy08ESC7bZJnxV3d/YQ5valTxBbzku/dQbXVj/xg69H8l8MxggGtMIIB
qQIBATCBgTB0MQswCQYDVQQGEwJVUzETMBEGA1UECBMKQ2FsaWZvcm5pYTEWMBQGA1UEBxMNTW91
bnRhaW4gVmlldzEUMBIGA1UEChMLR29vZ2xlIEluYy4xEDAOBgNVBAsTB0FuZHJvaWQxEDAOBgNV
BAMTB3dlYnZpZXcCCQDYug4AUBuGUzANBglghkgBZQMEAgEFADANBgkqhkiG9w0BAQEFAASCAQCw
pWR1Q0H9zy/vWmBqmHghBdGIZi29aQg/xtZ8G2RKjdp0rxIQfG/q7FttCBE9tY5aMD2PoNFDsZ7K
Z7BS22IJte9EhtREESaBEqRbYpk0AJX/aC35nhhpeadbRj/pllS5VmWU9yzHnNZkakP8cB+L4f6M
6sPg5rMM6ZWeXEkkdMZHdKDTml10A8rc3R4H21XWBMw1lcnaoehzka6EZtWWTsW3d8psRKojyq++
NNR5e2eqtrmGLpC+J2pp4va6t5FVy9+RbNU4WuCiObVp1zQex5f8BkLK8OYEb3d3EbQbeIQBMcfo
60as+OZlHdYDgU9TWnp4KYhgBvX91dp9XddI
```

rapperskull:
Other findings:
You can't have more than one package with _isFallback_ set to true. If two are found, a runtime exception is thrown.
You can't have a package with _isFallback_ set to true, but _availableByDefault_ set to false (or not set at all), otherwise a runtime exception is thrown.
_There must be at least one WebView package that is available by default_
_There must be at least one WebView package that is available by default and not a fallback_
The last two are self explanatory log messages in the code.

Based on this, we can tell how a patch should be done:
1. It is not necessary to declare a fallback package, which is good for us, since only one can be declared, but we don't know which one between the AOSP WebView, the Google WebView and Chrome is installed.
2. It is possible to declare more than one _availableByDefault_ package. If no package is selected, or the choice is not valid (eg. the user uninstalled the previous selected package), it fallback to the first installed _availableByDefault_ package. So basically the only package that should always be there is the fallback one (which fortunately is not mandatory to declare).

mfonville:
The whole purpose of the *fallback* flag seems to be that a package marked as such will be kept in a disabled state if a 'better' choice is available https://github.com/android/platform_frameworks_base/blob/nougat-release/services/core/java/com/android/server/webkit/WebViewUpdateServiceImpl.java#L133

Blystad:
So the signature needs to be updated whenever the Chrome APK is updated?
Excellent...just excellent...

I guess we should create a script that generates this file for us.

mfonville:
@Blystad I have already scripted something to extract the signature easily: https://github.com/opengapps/opengapps/blob/master/show_apksignature.sh
But the signature should be relatively static for each variant.
Especially now that they included themselves (at least for Dev/Beta/Canary) these signatures they won't be able to easily change them anymore.

So I think we can probably assume that all e.g. all Chrome arm64 sdk24 APKs will retain the same signature overtime.

Blystad:
Ah, it's just the certificate...Good, yeah, that won't change.

rapperskull:
Further informations about the signatures:
Thanks to @mfonville we discovered that it is not the actual base64 hash of the certificate that gets compared, but the single fields. In fact you can have two completely identical certificates but with different base64 encoding, because of small binary differences that do not affect the certificate verification process. Based on this information, I compared the actual certificates (not the hashes) and found out that there's no difference between architectures, which means easier to maintain patches.
BUT there are still some things to point out: the only valid certificate (for Chrome Stable, Beta and Dev) is the one that comes with the APK with minsdkversion set to 24 (Android 7.0), which is different from the one in the package with min API set to 21.
Also the certificates of Chrome Beta and Dev, and Google WebView are RSA SHA1, while they previously were RSA MD5, so they also changed the certificate to implement a more secure encryption.
Unfortunately Chrome Stable still uses an RSA MD5 certificate. This is bad for two things:
1. It's less secure (https://en.wikipedia.org/wiki/MD5#Security), but if Google is not concerned, why should we?
2. Since the certificate is not stored in the official XML file from Google, they could change it at any time, both for switching to SHA1 that for no reasons.
Chrome 53 Stable has already been released for Desktop, and should arrive on Android soon. My guess is that they will change the certificate and stick with it for a while, but it's just my thought.
We now have to wait and see what happens.

mfonville:
We also have btw still this issue that was reported: https://github.com/opengapps/opengapps/issues/348
are those permissions granted in a new permission patch?

Blystad:
I don't use the MM permission patch. It's unnecessary if the ROM is configured correctly (which entails having the overlays that we have here in the `aosp_build` project.)

Just removing the `Provision` package ( ref: https://github.com/opengapps/aosp_build/pull/76/commits/247b5f5a45b0387e3a91800a78698694e1346d6f ) was enough for me to have the SetupWizard be granted the correct permissions.

mfonville:
ok, thanks.

another request, I got word from mrBanks that  `GoogleExtShared  GoogleExtServices GooglePrintRecommendService` all three would have already be bundled in the AOSP code.
Is that correct? (and do they share the same packagename with `com.google.android.*`?)

alviteri:
@Blystad if you don't merge this then roms have signature errors installing via recovery .... Can you merge and search for a fix later for webview?

Blystad:
@mfonville as far as I can see, those sources aren't part of the normal AOSP nougat release. I tried to grep from them here.

@alviteri Sure

alviteri:
Thanks @Blystad  .... Did you check my PR?

mfonville:
@Blystad I got more info from mrBanks that their name should be e.g. `PrintRecommendService` (so without the `Google` part) on AOSP.
According to beanstown ROM dev they should be technically equivalent though. But it might need a 'replace' in the overlay.

rapperskull:
Chrome 53 for Android is out, and the signature for Chrome has not changed (yet). Fortunately this means that ROMs don't have to be updated, but also leaves us with the doubt that they could change it at any time.

osm0sis:
Hey guys, looks like there's an issue with Chrome v55. Flashing Open GApps 7.1 Stock 20161208 removed Chrome Stable Webview (which I actually originally noticed when the Open GApps app wouldn't open; Chrome was my only installed Webview in my current setup). Chrome the app itself appeared to be installed and otherwise working fine.

Downgrading back to my previous downloaded GApps from 20161205 resolved the issue and returned Chrome Stable Webview to working order. So I'm guessing v55 may have resulted in a signature change?

KreAch3R:
Hello guys, 

I've been researching the new Webview code since Nougat dropped, because I was still having intermittent issues with this patch and the ROM would still have trouble locating valid providers under specific (but randomly looking) circumstances.

I had to resort to extensive debugging using this commit: [krexus/frameworks_base@036750e](https://github.com/krexus/frameworks_base/commit/036750e2c276e30876c99b806c5be36f73f2a07e)

but I think I found the core of my issues and it was twofold: Firstly, this **systemIsDebuggable()** call here: [L630](https://github.com/krexus/frameworks_base/commit/036750e2c276e30876c99b806c5be36f73f2a07e#diff-36aec002eeba125dcb9755bcab856507R634)

was very sneaky and I think that made a lot of you think that this patch is working, while it should **never check for signatures, if you build USERDEBUG builds**. 

Secondly, I found out that the used signatures in this patch didn't match with the packages flashed to my device (Nexus 5, hammerhead, arm, 7.0/7.1), so I had to find a way to extract the correct signatures, as I describe here: [krexus/vendor_krexus@ef4a14d](https://github.com/krexus/vendor_krexus/commit/ef4a14d2471933e1c8e4da72fde6fde6d87d69d8)

Last but not least, I found out that we can use **multiple** *signature* tags inside every webview *provider*  tag (I didn't actually test it on an arm64 device but I added my valid signatures as second ones so I could force a miss handling, and it didn't happen).

So, to sum up:

1. Most ROMs building userdebug builds were avoiding the signature test
2. Some could say that we can skip all this debugging and just make **providerHasValidSignature** always return **TRUE** (and stop worrying about signatures)
3. We can use multiple *signature* tags inside the .xml

@mfonville This was the issue I was telling you about, I'm glad I finally found the time to explain my findings to all of you. Looking forward to your thoughts.

mfonville:
@KreAch3R thanks for your extensive research into this issue;

Did your method to extract the signature deliver the same signature as that we did find in the decompiled framework from Google?

```
unzip "$1" "META-INF/CERT.RSA"
openssl pkcs7 -in "META-INF/CERT.RSA" -inform DER -print_certs -out cert.pem
tail -n +4 cert.pem | head -n -2 > cert.partial
tr -d '\n' < cert.partial > cert.txt
```

mfonville:
Ok, I now checked myself, trying to replicate the Chrome Beta signature, and it matches 100%! So great job @KreAch3R !

So the correct signature for the regular chrome should be instead of current value:
```
MIIEQzCCAyugAwIBAgIJAMLgh0ZkSjCNMA0GCSqGSIb3DQEBBAUAMHQxCzAJBgNVBAYTAlVTMRMwEQYDVQQIEwpDYWxpZm9ybmlhMRYwFAYDVQQHEw1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKEwtHb29nbGUgSW5jLjEQMA4GA1UECxMHQW5kcm9pZDEQMA4GA1UEAxMHQW5kcm9pZDAeFw0wODA4MjEyMzEzMzRaFw0zNjAxMDcyMzEzMzRaMHQxCzAJBgNVBAYTAlVTMRMwEQYDVQQIEwpDYWxpZm9ybmlhMRYwFAYDVQQHEw1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKEwtHb29nbGUgSW5jLjEQMA4GA1UECxMHQW5kcm9pZDEQMA4GA1UEAxMHQW5kcm9pZDCCASAwDQYJKoZIhvcNAQEBBQADggENADCCAQgCggEBAKtWLgDYO6IIrgqWbxJOKdoR8qtW0I9Y4sypEwPpt1TTcvZApxsdyxMJZ2JORland2qSGT2y5b+3JKkedxiLDmpHpDsz2WCbdxgxRczfey5YZnTJ4VZbH0xqWVW/8lGmPav5xVwnIiJS6HXk+BVKZF+JcWjAsb/GEuq/eFdpuzSqeYTcfi6idkyugwfYwXFU1+5fZKUaRKYCwkkFQVfcAs1fXA5V+++FGfvjJ/CxURaSxaBvGdGDhfXE28LWuT9ozCl5xw4Yq5OGazvV24mZVSoOO0yZ31j7kYvtwYK6NeADwbSxDdJEqO4k//0zOHKrUiGYXtqw/A0LFFtqoZKFjnkCAQOjgdkwgdYwHQYDVR0OBBYEFMd9jMIhF1Ylmn/Tgt9r45jk14alMIGmBgNVHSMEgZ4wgZuAFMd9jMIhF1Ylmn/Tgt9r45jk14aloXikdjB0MQswCQYDVQQGEwJVUzETMBEGA1UECBMKQ2FsaWZvcm5pYTEWMBQGA1UEBxMNTW91bnRhaW4gVmlldzEUMBIGA1UEChMLR29vZ2xlIEluYy4xEDAOBgNVBAsTB0FuZHJvaWQxEDAOBgNVBAMTB0FuZHJvaWSCCQDC4IdGZEowjTAMBgNVHRMEBTADAQH/MA0GCSqGSIb3DQEBBAUAA4IBAQBt0lLO74UwLDYKqs6Tm8/yzKkEu116FmH4rkaymUIE0P9KaMftGlMexFlaYjzmB2OxZyl6euNXEsQH8gjwyxCUKRJNexBiGcCEyj6z+a1fuHHvkiaai+KL8W1EyNmgjmyy8AW7P+LLlkR+ho5zEHatRbM/YAnqGcFh5iZBqpknHf1SKMXFh4dd239FJ1jWYfbMDMy3NS5CTMQ2XFI1MvcyUTdZPErjQfTbQe3aDQsQcafEQPD+nqActifKZ0Np0IS9L9kR/wbNvyz6ENwPiTrjV2KRkEjH78ZMcUQXg0L3BYHJ3lc69Vs5Ddf9uUGGMYldX3WfMBEmh/9iFBDAaTCK
```

KreAch3R:
@mfonville Thanks! Yes, indeed it matched Chrome Beta, this was exactly how I came up with the correct way to extract it (blindly testing how to produce the same string).

I didn't/hadn't checked the stock image framework-res.apk at all (I based my research on that your Chrome Beta/Chrome Canary seemed to always work for me).

According to my tests, yes, this Chrome signature works everywhere, regardless of /system installation (during flash) or Play Store installation (& in **USER** builds). Google Webview signature is also different (check my commit).

So the other signature is completely broken for everyone? I was not sure if I had to replace or append, thinking about the possibility that multiple signatures could exist across the different Chrome packages (API version, architecture etc), and that was how you came up with it.

At least now we know though that we should be able to append more.

mfonville:
And this is the webview (checked on arm, arm64; both regular and their beta):
```
MIIDuzCCAqOgAwIBAgIJANi6DgBQG4ZTMA0GCSqGSIb3DQEBBQUAMHQxCzAJBgNVBAYTAlVTMRMwEQYDVQQIDApDYWxpZm9ybmlhMRYwFAYDVQQHDA1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKDAtHb29nbGUgSW5jLjEQMA4GA1UECwwHQW5kcm9pZDEQMA4GA1UEAwwHd2VidmlldzAeFw0xNDA4MDgyMzIwMjBaFw00MTEyMjQyMzIwMjBaMHQxCzAJBgNVBAYTAlVTMRMwEQYDVQQIDApDYWxpZm9ybmlhMRYwFAYDVQQHDA1Nb3VudGFpbiBWaWV3MRQwEgYDVQQKDAtHb29nbGUgSW5jLjEQMA4GA1UECwwHQW5kcm9pZDEQMA4GA1UEAwwHd2VidmlldzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAMbtaFX0r5aZJMAbPVMAgK1ZZ29dTn91VsGxXv2hqrQo7IpqEy2JmPvPnoMsSiuTAe+UcQy8oKDQ2aYVSAd1DGIy+nSRyFTt3LSIAdwSBkB1qT4a+OqkpsR6bSNXQXQ18lCQu9gREY3h3QlYBQAyzRxw4hRGlrXAzuSz1Ec4W+6x4nLG5DG61MAMR8ClF9XSqbmGB3kyZ70A0X9OPYYxiMWP1ExaYvpaVqjyZZcrPwr+vtW8oCuGBUtHpBUH3OoG+9s2YMcgLG7vCK9awKDqlPcJSpIAAj6uGs4gORmkqxZRMskLSTWbhP4p+3Ap8jYzTVB6Y1/DMVmYTWRMcPW0macCAwEAAaNQME4wHQYDVR0OBBYEFJ6bAR6/QVm4w9LRSGQiaR5Rhp3TMB8GA1UdIwQYMBaAFJ6bAR6/QVm4w9LRSGQiaR5Rhp3TMAwGA1UdEwQFMAMBAf8wDQYJKoZIhvcNAQEFBQADggEBAEQu8QiVxax7/diEiJrgKE1LwdXsIygJK/KnaKdnYEkAQpeu/QmrLiycm+OFbL1qHJIB7OuI/PQBUtcaNSiJSCVgtwtEbZWWIdsynqG/Nf4aGOndXegSQNRH54M05sRHLoeRycPrY7xQlEwGikNFR76+5UdwFBQI3Gn22g6puJnVukQm/wXQ+ajoiS4QclrNlixoDQsZ4STLH4+Wju2wIWKFFArIhVEIlbamq+p6BghuzH3aIz/Fy0YTQKi7SA+0fuNeCaqlSm5pYSt6p5CH89y1Fr+wFc5r3iLRnUwRcy08ESC7bZJnxV3d/YQ5valTxBbzku/dQbXVj/xg69H8l8M
```

PR #77, opened 2016-09-01
Allow override AOSP DeskClock

alviteri:
also include needed overlay for frameworks


alviteri:
PR Updated @Blystad

PR #86, opened 2016-09-26
Fix the WebView overlay API 24 folder derp

AdFad666:
Now the overlay actually works!


Blystad:
Perfect. Thanks!

PR #90, opened 2016-09-29
Handle lz compressed apks

jamuir:
Some apks are lz compressed to workaround github file-size limits.
Use "lunzip" to decompress them.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -22,6 +22,14 @@ ifeq ($(GAPPS_FORCE_MATCHING_DPI),false)
   endif
 endif
 
+GAPPS_LUNZIP_REQUIRED := $(shell find $(GAPPS_SOURCES_PATH) -name '*.apk.lz' -print -quit)
+ifneq ($(GAPPS_LUNZIP_REQUIRED),)
+  GAPPS_TEST_LUNZIP := $(shell command -v lunzip)
+  ifeq ($(GAPPS_TEST_LUNZIP),)
+    $(error lunzip is not available. Please install it first ("sudo apt-get install lunzip"))
+  endif
+endif
+
 include $(GAPPS_BUILD_SYSTEM_PATH)/definitions.mk
 
 # Device should define their GAPPS_VARIANT in device/manufacturer/product/BoardConfig.mk
--- a/core/find_apk.sh
+++ b/core/find_apk.sh
@@ -8,7 +8,15 @@ getlatestapk() {
   tempapks=( )
   OLDIFS="$IFS"
   IFS="
-" # We set IFS to newline here so that spaces can survive the for loop
+" # We set IFS to newline here so that filenames containing spaces are
+  # handled correctly in the for-loops below.
+
+  # decompress apks
+  # some apks are lz compressed to work around github file-size limits.
+  for foundapklz in $(find "$1" -iname '*.apk.lz'); do
+    lunzip --keep "$foundapklz"
+  done
+
   # sed copies filename to the beginning, to compare version, and later we remove it with cut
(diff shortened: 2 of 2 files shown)


core/find_apk.sh
```diff
@@ -8,7 +8,15 @@ getlatestapk() {
...
   OLDIFS="$IFS"
   IFS="
-" # We set IFS to newline here so that spaces can survive the for loop
+" # We set IFS to newline here so that filenames containing spaces are
+  # handled correctly in the for-loops below.
+
+  # decompress apks
+  # some apks are lz compressed to workaround github file-size limits.
+  for foundapklz in $(find "$1" -iname '*.apk.lz'); do
+    lunzip $foundapklz
```
mfonville: I think you will have to quote "$foundapklz" to avoid issues if the path has spaces

Blystad: Will lunzip be a no-op if the file is already extracted?

Also, could we make sure that we clean up old extracted APKs? I would prefer not having to SSH into a build server and manually clean it up every few weeks :)

Blystad: Lets do something like
```
lunzip --keep "$foundapklz"
```

To avoid getting a "deleted" entry in git for the arm64 repo. That can lead to confusion. Also lunzip is no-op if the file is already extracted, so that'll be fine.

For cleaning up, maybe do something like:
```
for f in $(find -iname '*.apk'); do test -e $f.lz || rm $f; done     
```

config.mk
```diff
@@ -22,6 +22,11 @@ ifeq ($(GAPPS_FORCE_MATCHING_DPI),false)
   endif
 endif
 
+GAPPS_TEST_LUNZIP := $(shell command -v lunzip)
```
Blystad: Lets just require this for arm64 for now, just check

```
ifeq ($(TARGET_ARCH),arm64)
```

core/find_apk.sh
mfonville: You can check out the code at https://github.com/opengapps/opengapps/blob/master/download_sources.sh#L63-L69

it has the example commands for extraction and for clean-up

config.mk
Blystad: Fair enough. Since config.mk is only loaded once during build...maybe we can make it smart? Do something like

```
GAPPS_LZ_REQUIRED := $(shell find . -name '*.apk.lz' -print -quit)
ifneq ($(GAPPS_LZ_REQUIRED),)
// if not lunzip is installed, throw error
endif
```

core/find_apk.sh
jamuir: @mfonville:  agreed.  it is better to use quotes.  change made.

@Blystad: I've changed the command to "lunzip --keep".

**re clean-up.**  This is important.  If "foo.apk.lz" is updated in a subsequent commit, then "foo.apk" won't be regenerated if that file already exists.  However, I don't think the clean-up code can go inside the script "find_apk.sh".

Ideally, the decompressed apks would reside under $OUT, and then they would be removed when "make clean" or "make installclean" is run.  But that would likely require a major refactor of "find_apk.sh".

I will think more about the clean-up, but I'd like to defer that to later PR.

If the current PR is merged, then aosp_build will start working on bullhead.

config.mk
jamuir: I like it, but we need to use the right dir.

Maybe this:
```
GAPPS_LZ_REQUIRED := $(shell find vendor/opengapps/sources -name '*.apk.lz' -print -quit)
```

jamuir: I've updated the branch with this change.

core/find_apk.sh
Blystad: Regarding clean up, since the APK name always changes (since it's equivalent to the APK version number), that shouldn't be a problem.

But, rest of the PR looks good to me, and I'll create a task to clean up extracted .lz files.

Thanks for your contribution!

Issue #92, opened 2016-09-30
Frameworks now also API-dependent


mfonville:
Sorry for creating more work, but because Nougat frameworks seem to be (at least partly) binary incompatible with Marshmallow, frameworks are now stored in a different directory per API.

See this change in opengapps's buildcode:
https://github.com/opengapps/opengapps/commit/134a7b66caf66f9c636f1b6444a3a95aa377c320
and these commits in source/all:
https://github.com/opengapps/all/compare/81778571e3...da2d949988

Similar changes will have to be applied to aosp_build to deal with this new layout

Blystad:
Well isn't that great :) Can you do me a favour and just symlink 20, 21, 22 and 23 to 19? so I don't need to loop through them all. Makes my life a lot easier.

mfonville:
It is based on the library-finding code, wouldn't it be possible to adjust that (I will take a small peek, to see if I can suggest the exact lines you'd need)

Blystad:
The framework copying is done in `opengapps-files.mk`, which uses some makefile definitions from `core/definitions.mk` to find all files in the specific folders, and convert them into the structure

```
vendor/opengapps/sources/all/framework/...:system/framework/...
```

so that PRODUCT_COPY_FILES get what it expects. We don't use the find_apk.sh script for this at all.

mfonville:
ok.
I am btw not sure if symlinking would do the trick, e.g. for camera2 framework I don't have a Nougat version, so the Nougat packages fallback to the pre-Nougat version for that one.

Blystad:
Ok...In that case:

Can you do me a favour and provide the result of `find sources/all -type f`? I'm on a shit internet connection, so I'm not going to sync `sources/all` for at least a couple of days. Then I can write a script that handles this in a proper way.

mfonville:
ofc np, here you go: 
```
sources/all/.gitignore
sources/all/HOWTO
sources/all/LICENSE
sources/all/usr/srec/en-US/endpointer_model
sources/all/usr/srec/en-US/lstm_model.uint8.data
sources/all/usr/srec/en-US/metadata
sources/all/usr/srec/en-US/c_fst
sources/all/usr/srec/en-US/grammar.config
sources/all/usr/srec/en-US/lexicon.U.fst
sources/all/usr/srec/en-US/ep_portable_mean_stddev
sources/all/usr/srec/en-US/voice_actions.config
sources/all/usr/srec/en-US/CONTACT_NAME.syms
[...]

PR #94, opened 2016-10-05
Support API dependent copyable resources

Blystad:
From Android N we need platform API dependent system/framework.

Fixes issue #92

Change-Id: Iabe04e2a3e78537107d4a761712ae5d8e36f1b71


Blystad:
Script updated to pass pylint with no warnings.

PR #96, opened 2016-10-10
aosp: use better code style

alviteri:

### DIFF
--- a/opengapps-packages.mk
+++ b/opengapps-packages.mk
@@ -1,100 +1,111 @@
 include vendor/google/build/config.mk
 include $(GAPPS_FILES)
 
-DEVICE_PACKAGE_OVERLAYS += $(GAPPS_DEVICE_FILES_PATH)/overlay/pico
+DEVICE_PACKAGE_OVERLAYS += \
+    $(GAPPS_DEVICE_FILES_PATH)/overlay/pico
 
-GAPPS_PRODUCT_PACKAGES += GoogleBackupTransport \
-                    GoogleContactsSyncAdapter \
-                    GoogleFeedback \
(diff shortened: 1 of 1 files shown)


jamuir:
thanks for cleaning up the makefile :-)

those long multi-line variable assignments are much easier to read now.

Blystad:
I've moved the GAPPS_PRODUCT_PACKAGES filter to the end of the file in commit https://github.com/opengapps/aosp_build/commit/cb3629afdb53bca92ae5ba1b9f75f6788b7a0c0d

so it should be fine now.

jamuir:
@alviteri , @Blystad:  I should have given more explanation about why the filter command wasn't placed at the end of the file.

The problem is that the modules considered below line 117 also update DEVICE_PACKAGE_OVERLAYS.

Consider someone who is using `stock` but doesn't want `Chrome` installed.  Then they would add `Chrome` to GAPPS_EXCLUDED_PACKAGES.

In that case, we should not update `DEVICE_PACKAGE_OVERLAYS` for the Chrome overlay, but with the filter command at the end of the end, that is what will happen.

joint filtering of DEVICE_PACKAGE_OVERLAYS and PRODUCT_PACKAGES seemed complicated to me, so I avoided it by not placing `PRODUCT_PACKAGES += $(filter-out $(GAPPS_EXCLUDED_PACKAGES),$(GAPPS_PRODUCT_PACKAGES))` at the end of the file.  I should made this more obvious by adding a comment above the `filter-out` command.

Blystad:
Right....good point. I didn't consider that. Let's get that fixed. If
you're up for fixing it I will approve the PR tonight. Otherwise I'll fix
it myself later tonight or tomorrow :)

Issue #91, opened 2016-09-29
Cleanup unnecessary extracted .lz files


Blystad:
We now extract .lz files and keep them for old APKs. We'll properly select the new APK, but we should delete the old APKs, otherwise the vendor/opengapps/sources dirs will grow for each new updated version.

See PR #90 for a discussion on this topic.

jamuir:
a clean-up operation is implemented in PR https://github.com/opengapps/aosp_build/pull/97.

PR #97, opened 2016-10-11
Clean-up decompressed apks

jamuir:
Some apks are lz compressed and then are decompressed under
$(GAPPS_SOURCES_PATH).  Add a clean-up step to get rid of decompressed
files.


jamuir:
this addresses Issue https://github.com/opengapps/aosp_build/issues/91

PR #99, opened 2016-10-14
Fix GAPPS_FORCE_* product packages

Blystad:


Blystad:
@jamuir please see this.

jamuir:
Looks good to me.  Thanks for fixing this!

opengapps-packages.mk
```diff
@@ -149,18 +155,16 @@ ifeq ($(GAPPS_FORCE_MMS_OVERRIDES),true)
 DEVICE_PACKAGE_OVERLAYS += \
     $(GAPPS_DEVICE_FILES_PATH)/overlay/mms
 
-GAPPS_PRODUCT_PACKAGES += \
+PRODUCT_PACKAGES += \
     PrebuiltBugle
 endif
 
 ifeq ($(GAPPS_FORCE_PIXEL_HOME),true)
 GAPPS_EXCLUDED_PACKAGES += \
```
jamuir: this update to `GAPPS_EXCLUDED_PACKAGES` was added in https://github.com/opengapps/aosp_build/commit/20ba41b8e5d326104f3f276ff9ef96f1e9804c67

it doesn't actually do anything because `GAPPS_EXCLUDED_PACKAGES` isn't referenced anywhere else.  To avoid confusion, it could be dropped (but, it isn't causing any harm right now).

PR #100, opened 2016-10-23
build: Fix for Pixel Launcher

alviteri:
make PixelLauncher a PRIVILEGED APP as https://github.com/opengapps/all/tree/master/priv-app/com.google.android.apps.nexuslauncher/21/nodpi
make PixelLauncherIcons builable for SDK >= 25 (Android Nougat 7.1) in old SDK it causes errors in build
rename Wallpaper to Wallpapers
match Google Play names

### DIFF
--- a/opengapps-packages.mk
+++ b/opengapps-packages.mk
@@ -159,12 +159,16 @@ PRODUCT_PACKAGES += \
     PrebuiltBugle
 endif
 
-ifeq ($(GAPPS_FORCE_PIXEL_HOME),true)
+ifeq ($(GAPPS_FORCE_PIXEL_LAUNCHER),true)
 GAPPS_EXCLUDED_PACKAGES += \
     GoogleHome
 
+ifneq ($(filter $(call get-allowed-api-levels),25),)
 PRODUCT_PACKAGES += \
-    PixelHome \
-    PixelHomeIcons \
-    Wallpaper        
+    PixelLauncherIcons
+endif
+
+PRODUCT_PACKAGES += \
+    PixelLauncher \
+    Wallpapers
 endif
--- a/README.md
+++ b/README.md
@@ -102,10 +102,10 @@ If you want to include Chrome on a non-full build you need:
 GAPPS_FORCE_BROWSER_OVERRIDES := true
 ```
 
-If you want use PixelHome overriding GoogleHome you need:
+If you want use PixelLauncher overriding GoogleHome you need:
 
 ```
-GAPPS_FORCE_PIXEL_HOME := true
+GAPPS_FORCE_PIXEL_LAUNCHER := true
 ```
 
 On a per-app basis, add the GApps package to `GAPPS_PACKAGE_OVERRIDES`.
--- a/modules/PixelLauncher/Android.mk
+++ b/modules/PixelLauncher/Android.mk
@@ -1,8 +1,9 @@
 LOCAL_PATH := .
 include $(CLEAR_VARS)
 include $(GAPPS_CLEAR_VARS)
-LOCAL_MODULE := PixelHome
+LOCAL_MODULE := PixelLauncher
 LOCAL_PACKAGE_NAME := com.google.android.apps.nexuslauncher
+LOCAL_PRIVILEGED_MODULE := true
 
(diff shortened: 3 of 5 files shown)


alviteri:
@Blystad DO NOT MERGE until test on M .... for remove the api-level conditional ... building now on M here

alviteri:
@Blystad if the changes are ok for you ... you can merge this.... I can't test on L ..... because I am not building L anymore ..... so I put the conditional for make this buildable from M...

screenshot_20161023-134317

mfonville:
@alviteri the minsdk entry in the APK is also >=21 so MM seems the valid point for the conditional

mfonville:
I just wonder a bit, why is the keyword rephrased to "Nexus Laucher" while Google puts every human-readable name as "Pixel Launcher"?

alviteri:
ok .... I just want follow package name .... I will let pixelhome ..... about the launcher you mean this is ok for L too?

mfonville:
Yeah, sorry, my mistake, I wrote MM, but meant Lollipop actually X-D
I did not test, but if minSDK works, for both the PixelLauncher (sdk21) as Wallpapers (sdk16) I think it should work! :-)

alviteri:
ok man let me do the changes

alviteri:
just one more question .... is this able to build KK or from L only? I mean for add or not the conditional for L (api21)

alviteri:
@mfonville check now please

mfonville:
Could your rename Wallpaper to Wallpaper**s**? That would bring it in line with the PlayStore and OpenGApps packages keyword.

Also using "PixelLauncher" instead of "PixelHome" would fit Google's use and the package-use.

Not sure about KK support of aosp_build.

Also on all the above, it is @Blystad's decision in the end, he is the lead dev on aosp_build.

alviteri:
@mfonville check now

opengapps-packages.mk
```diff
@@ -159,12 +159,19 @@ PRODUCT_PACKAGES += \
...
+ifeq (1,$(filter 1,$(shell echo "$$(( $(PLATFORM_SDK_VERSION) >= 25 ))" )))
+# never dexopt the PixelLauncherIcons
+$(call add-product-dex-preopt-module-config,PixelLauncherIcons,disable)
+
+PRODUCT_PACKAGES += \
+    PixelLauncherIcons
+endif
+
 PRODUCT_PACKAGES += \
     PixelHome \
```
mfonville: PixelHome still need to rennamed here to PixelLauncher

alviteri: thanks

alviteri:
fixed

mfonville:
LGTM, so we will now leave it up to @Blystad

opengapps-packages.mk
```diff
@@ -159,12 +159,19 @@ PRODUCT_PACKAGES += \
     PrebuiltBugle
 endif
 
-ifeq ($(GAPPS_FORCE_PIXEL_HOME),true)
+ifeq ($(GAPPS_FORCE_PIXEL_LAUNCHER),true)
 GAPPS_EXCLUDED_PACKAGES += \
     GoogleHome
 
+ifeq (1,$(filter 1,$(shell echo "$$(( $(PLATFORM_SDK_VERSION) >= 25 ))" )))
```
Blystad: Why can't this just use

```
ifneq ($(filter $(call get-allowed-api-levels),25),)
```

Like everything else?

opengapps-packages.mk
```diff
@@ -159,12 +159,19 @@ PRODUCT_PACKAGES += \
...
+# never dexopt the PixelLauncherIcons
+$(call add-product-dex-preopt-module-config,PixelLauncherIcons,disable)
```
Blystad: You can specify this in the PixelLauncherIcons package. See:

https://source.android.com/devices/tech/dalvik/configure.html#local_dex_preopt

alviteri:
Updated @Blystad

mfonville:
After some time, we now noticed from feedback of Lollipop users that PixelLauncher uses an (abstracted) API method that is only SDK23+, so though even it targets Lollipop with the APK, it seems to be Marshmallow+ only

Issue #101, opened 2016-10-31
SetupWizard is not set correctly for SDK 19


niks255:
On kitkat the correct name is com.google.android.setupwizard, not com.google.android.setupwizard.default. This should be fixed.

Blystad:
Should be fixed with #110

PR #105, opened 2016-11-25
package name of SetupWizard is corrected

moku:
Due to incorrect package name of SetupWizard module, compile fails.
Package name changed from 'com.google.android.setupwizard.default' to 'com.google.android.setupwizard'

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,10 +102,10 @@ If you want to include Chrome on a non-full build you need:
 GAPPS_FORCE_BROWSER_OVERRIDES := true
 ```
 
-If you want use PixelHome overriding GoogleHome you need:
+If you want use PixelLauncher overriding GoogleHome you need:
 
 ```
-GAPPS_FORCE_PIXEL_HOME := true
+GAPPS_FORCE_PIXEL_LAUNCHER := true
 ```
 
 On a per-app basis, add the GApps package to `GAPPS_PACKAGE_OVERRIDES`.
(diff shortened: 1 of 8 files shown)


mfonville:
@moku on 5.0+ the setupwizard is on purpose renamed with .default in the Open GApps sources repos, to differentiate the 'tablet' and the 'default' setupwizard (they are identified as such by decompiling the APK and looking at a value)

Blystad:
See issue #101 

It needs to be `.default` on 5.0+

You can do something like:

```
ifneq ($(filter $(call get-allowed-api-levels),21),)
# kitkat here
LOCAL_PACKAGE_NAME := com.google.android.setupwizard
else
# original code here, that handles LP and newer.
endif
```

also delete the .swp file.

moku:
Thanks for review.
I'll apply the comments and commit.

Blystad:
I think the .tablet version doesn't exist for KitKat, so you need to overwrite that as well for KitKat.

Do something like:

```
ifeq ($(filter $(call get-allowed-api-levels),21),)
    # Kitkat
    LOCAL_PACKAGE_NAME := com.google.android.setupwizard
else
    ifneq ($(filter $(PRODUCT_CHARACTERISTICS),tablet),)
        LOCAL_PACKAGE_NAME := com.google.android.setupwizard.tablet
    else
        LOCAL_PACKAGE_NAME := com.google.android.setupwizard.default
    endif
endif
```

Also, please squash your commits. :)

PR #110, opened 2016-12-09
support compatibility for kitkat

moku:
In kk, shared_libraries of prebuilt_apk are required to be handled manually.
In this commit, *.so files in prebuilt apk are extracted, added to LOCAL_SHARED_LIBRARIES, and compiled automatically.
Also, the package name of SetupWizard is com.google.android.setupwizard in only kitkat

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ modules/product.mk
 modules/product_copy_files.txt
 modules/app_densities.txt
 modules/README
+
+# makefile for shared library for kitkat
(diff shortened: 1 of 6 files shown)


Blystad:
I don't have a KitKat build to test, but I trust you did enough testing, and it doesn't seem like this will interfare with LP or newer, so I'll merge it.

niks255:
The problem with this is that libraries are duplicated inside the apk and in /system/lib. You probably need to handle that too.

PR #117, opened 2017-01-08
Only build gapps if a variant was specified

invisiblek:
* This can be useful if building for several devices in the same tree, but
  only wanting to supply built-in gapps for specific ones.

* Throw up an info message that gapps are being skipped if GAPPS_VARIANT
  was not specified, but don't bail on the build completely because of it.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -1,3 +1,5 @@
+ifneq ($(GAPPS_VARIANT),)
+
 # Opengapps AOSP build system
 GAPPS_BUILD_SYSTEM_PATH := vendor/google/build/core
 GAPPS_SOURCES_PATH := vendor/opengapps/sources
@@ -37,9 +39,6 @@ endif
 include $(GAPPS_BUILD_SYSTEM_PATH)/definitions.mk
 
 # Device should define their GAPPS_VARIANT in device/manufacturer/product/BoardConfig.mk
-ifeq ($(GAPPS_VARIANT),)
-  $(error GAPPS_VARIANT must be configured)
-endif
 GAPPS_VARIANT_EVAL := $(call get-gapps-variant,$(GAPPS_VARIANT))
 
 ifeq ($(GAPPS_VARIANT_EVAL),)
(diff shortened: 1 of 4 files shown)


Blystad:
Looks good, but I think this will still apply the `pico` device overlays.

```
DEVICE_PACKAGE_OVERLAYS += \
    $(GAPPS_DEVICE_FILES_PATH)/overlay/pico
```

as well as try to build, but fail to find

```
GAPPS_PRODUCT_PACKAGES += \
    GoogleBackupTransport \
    GoogleContactsSyncAdapter \
    GoogleFeedback \
    GoogleOneTimeInitializer \
    GooglePartnerSetup \
    PrebuiltGmsCore \
    GoogleServicesFramework \
    GoogleLoginService \
    SetupWizard \
    Phonesky \
    GoogleCalendarSyncAdapter
```

Maybe add a

```
ifneq ($(GAPPS_VARIANT),)
```

in `opengapps-packages.mk` at the top.

I don't think you need it in `clear_vars.mk` or `definitions.mk`

invisiblek:
Well, clear_vars.mk and definitions.mk get included automatically by the build system. In my opinion it'd be better to ignore anything in them if we're not building gapps. With the thought being that some day one could contain something that would break a build for a device that's not using them.

As for opengapps-packages.mk, since that's included manually in addition to setting the GAPPS_VARIANT, and not included automatically via the build system, I don't see why it makes a difference. (Unless we're looking to change the documentation to include that somewhere other than "`device/manufacturer/product/device.mk`")

Blystad:
valid point. New suggestion: Add a error if a user includes opengapps-packages.mk without setting GAPPS_VARIANT.

invisiblek:
Good idea. Additionally I think I want to remove the info warning if GAPPS_VARIANT wasn't set (at the end of config.mk). Its mildly annoying to see it during a clobber. Thoughts on that?

invisiblek:
Amended with the discussed items.

PR #120, opened 2017-02-09
Mark all the apks as presigned

gmrt:


gmrt:
This was done with
```
find -name Android.mk -exec sed -i '/LOCAL_PACKAGE_NAME.*/a LOCAL_CERTIFICATE := PRESIGNED' {} \;
```
so it's possible that I missed something, I had to fix just a couple of modules.

Dudulle:
This flag is already set actually
https://github.com/opengapps/aosp_build/blob/master/core/prebuilt_apk.mk#L4

gmrt:
I doubt that flag works since every mk includes clear_vars.mk, which clears LOCAL_CERTIFICATE. In fact, I couldn't complete my build until I did this.

Dudulle:
prebuilt_apk.mk is included at the end of every module.
The logic to locate the correct .apk and libs are inside that file so it shouldn't be possible to make it build without LOCAL_CERTIFICATE being set too.

Most likely there was another issue.

gmrt:
Yes, sorry, I noticed that soon after I replied. I suspect the issue was caused by not running make clean before starting a build with gapps, probably the apks were included, but not the info needed for the signing phase.

Issue #109, opened 2016-12-07
Setup BackupTransport overlay


Blystad:
See: https://android.googlesource.com/device/sample/+/4b3880d/overlays/backup/frameworks/base/packages/SettingsProvider/res/values/defaults.xml

We need to add that to make the Google Backup system work.

mfonville:
Should these also added by regular ROMs that want GApps compatiblity?

Blystad:
Yes. The default backup transport is meant for local testing.

Blystad:
Should we consider creating a repo with all overlays that is necessary? I've probably received 400 emails with people cherry-picking in the webview commit. Would be easier if we just told them

"Add this repo, and add this line to the end of your `device.mk` and you're good to go!"

mfonville:
@thecrazylex what is your opinion on an overlay repo?

TheCrazyLex:
I think the Version specific problems are not always related to overlays. Some may require in-code changes. So such a seperate repo wouldn't be a full solution.

Blystad:
Some things would be constant. E.g. the webview patch, the location provider and the backup provider.

If the location provider is configured correctly, then the marshmallow permission code patch is totally unnecessary (as is evident that it's not present if you decompile the framework.jar from an official system image).

But, I guess some patches are needed if you want to support Google Play opt-in. Instead of building it as part of the system image (as I do).

PR #121, opened 2017-02-10
aosp_build: correct path for backup service overlay

alviteri:
it is defined at defaults.xml since android-5.0.0_r1 tag

https://android.googlesource.com/platform/frameworks/base/+/android-5.0.0_r1/packages/SettingsProvider/res/values/defaults.xml#55


Blystad:
That should, strictly speaking, not matter at all.

All the XML files in res/values/X.xml gets merged together and compiled. At least that's how I understand the system.

Tried printing out the value seen in the Java/C++ layer and checked that it's correct?

alviteri:
I don't how much influent this but I have had this https://github.com/SonyAosp/device_sony_common/blob/android-7.1/overlay/frameworks/base/packages/SettingsProvider/res/values/defaults.xml and never seen the bug explained here https://github.com/opengapps/aosp_build/issues/109

alviteri:
https://github.com/LineageOS/android_vendor_cm/blob/cm-14.1/overlay/common/frameworks/base/packages/SettingsProvider/res/values/defaults.xml

Blystad:
Since this hasn't been updated since February I'm closing it.

PR #125, opened 2017-04-17
FaceLock Module : add condition to check if built target is arm arch

peterlolty:
The FaceLock module only support arm based arch, a condition added for checking arch to avoid build error on x86/x86_64 target.

### DIFF
--- a/opengapps-packages.mk
+++ b/opengapps-packages.mk
@@ -28,8 +28,12 @@ GAPPS_PRODUCT_PACKAGES += \
 endif
 
 ifneq ($(filter $(TARGET_GAPPS_VARIANT),nano),) # require at least nano
+# support only arm based platforms
+ifneq ($(filter arm%, $(TARGET_ARCH)),)
 GAPPS_PRODUCT_PACKAGES += \
-    FaceLock \
+    FaceLock
+endif
+GAPPS_PRODUCT_PACKAGES += \
     Velvet
 
 ifneq ($(filter $(TARGET_GAPPS_VARIANT),micro),) # require at least micro


Blystad:
Doesn't x86 typically ship with a ARM compatibility layer?

Would "facelock" work on X86, or would that be broken by this commit?

mfonville:
In general since lollipop the x86 targets also support arm with some hardware tricks (it can emulate at a low level), but since Facelock uses those jni lsystem libs that are linked to other libs too, like libfrsdk, it might be that they don't work on x86.

peterlolty:
I just follow the logic from "opengapps/scripts/inc.buildtarget.sh" which explicitly exclude the Facelock module for non-arm arch. Btw, Asus Zenfone 2 is a x86 device with arm translation support, but it still disable the Facelock module, so it should be a difficulty to run on x86 devices.

Dudulle:
In this case, I think it would be cleaner to put the condition in opengapps-packages.mk instead of in the module directly.

peterlolty:
@Dudulle It seems a proper place to put the logic there, I have push a new commit.

opengapps-packages.mk
```diff
@@ -28,8 +28,12 @@ GAPPS_PRODUCT_PACKAGES += \
 endif
 
 ifneq ($(filter $(TARGET_GAPPS_VARIANT),nano),) # require at least nano
+# support only arm based platforms
+ifeq ($(TARGET_ARCH), $(filter $(TARGET_ARCH), arm arm64))
```
jamuir: https://github.com/opengapps/opengapps/blob/master/scripts/inc.buildtarget.sh#L280

The following check emulates the logic in `inc.buildtarget.sh` more closely:

```makefile
ifneq ($(filter arm%,$(TARGET_ARCH)),) 
...
endif
```
But, this is a minor suggestion, so feel free to ignore it :-)

peterlolty: Good to learn from that, thank you.

opengapps-packages.mk
```diff
@@ -28,8 +28,12 @@ GAPPS_PRODUCT_PACKAGES += \
 endif
 
 ifneq ($(filter $(TARGET_GAPPS_VARIANT),nano),) # require at least nano
+# support only arm based platforms
+ifneq ($(filter $(TARGET_ARCH), arm%),)
```
jamuir: I think you want this instead:

```makefile
ifneq ($(filter arm%,$(TARGET_ARCH)),)
```

peterlolty: oh, thank you, i didn't take care of the %, let me push a new commit

peterlolty:
Is there any problem about this pull request? So when would you guys merge this?

Blystad:
Can you rebase it on top of master, so we avoid the merge commit?

Then I'll include it.

peterlolty:
@Blystad Please help to merge, thank you.

PR #129, opened 2017-04-17
Start using the path "vendor/opengapps/build"

jamuir:
The "aosp_build" project was previously installed under the path
"vendor/google/build".  This was done to take advantage of a
google-specific include statement in build/core/main.mk:

```makefile
# Include the google-specific config
-include vendor/google/build/config.mk
```
However, this caused problems because it meant that the opengapps
build files were always processed, even when building for a device
that should not have opengapps added to it (this was the reason for
Dan P's change, da56d8b0); it also caused some generic makefile
targets like "make clean" to break (unless GAPPS_VARIANT was set).

Start using the path "vendor/opengapps/build" for "aosp_build" and
avoid issues related to unwanted processing of "config.mk".

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -1,11 +1,30 @@
-ifneq ($(GAPPS_VARIANT),)
-
 # Opengapps AOSP build system
-GAPPS_BUILD_SYSTEM_PATH := vendor/google/build/core
+GAPPS_BUILD_SYSTEM_PATH := vendor/opengapps/build/core
 GAPPS_SOURCES_PATH := vendor/opengapps/sources
-GAPPS_DEVICE_FILES_PATH := vendor/google/build
-GAPPS_FILES := $(GAPPS_DEVICE_FILES_PATH)/opengapps-files.mk
+GAPPS_DEVICE_FILES_PATH := vendor/opengapps/build
 GAPPS_CLEAR_VARS := $(GAPPS_BUILD_SYSTEM_PATH)/clear_vars.mk
+
+BUILD_GAPPS_PREBUILT_APK := $(GAPPS_BUILD_SYSTEM_PATH)/prebuilt_apk.mk
+BUILD_GAPPS_PREBUILT_SHARED_LIBRARY := $(GAPPS_BUILD_SYSTEM_PATH)/prebuilt_shared_library.mk
+
+# check that we reside under the expected path, otherwise the
+# variables defined above are invalid
+ifeq ($(wildcard $(GAPPS_DEVICE_FILES_PATH)/config.mk),)
+  $(error Please update your manifest to use the path "$(GAPPS_DEVICE_FILES_PATH)" for the "aosp_build" project)
+endif
+
+ifeq ($(GAPPS_VARIANT),)
+  $(error GAPPS_VARIANT must be configured)
+endif
+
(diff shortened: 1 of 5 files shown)


jamuir:
@mfonville :  this is a fairly significant refactor.  I've tested various build commands (`mm`, `mmm`) and they seem to be working correctly, but it would be good to get a second opinion.

If you wanted to try this out, then you will need to update your manifest and replace the path `vendor/google/build` with `vendor/opengapps/build`.

config.mk
```diff
@@ -35,16 +54,3 @@ ifneq ($(GAPPS_LUNZIP_REQUIRED),)
...
-# Device should define their GAPPS_VARIANT in device/manufacturer/product/BoardConfig.mk
-GAPPS_VARIANT_EVAL := $(call get-gapps-variant,$(GAPPS_VARIANT))
-
-ifeq ($(GAPPS_VARIANT_EVAL),)
-  $(error GAPPS_VARIANT $(GAPPS_VARIANT) was not found. Use of one of pico,nano,micro,mini,full,stock,super)
-endif
-
-TARGET_GAPPS_VARIANT := $(GAPPS_VARIANT_EVAL)
-
-endif
```
jamuir: `GAPPS_VARIANT` checks were moved near the top of this file.

config.mk
```diff
@@ -1,12 +1,31 @@
-ifneq ($(GAPPS_VARIANT),)
-
 # Opengapps AOSP build system
-GAPPS_BUILD_SYSTEM_PATH := vendor/google/build/core
+GAPPS_BUILD_SYSTEM_PATH := vendor/opengapps/build/core
```
Blystad: If we change this to:

```
GAPPS_DEVICE_FILES_PATH := $(call my-dir)
GAPPS_BUILD_SYSTEM_PATH := $(GAPPS_DEVICE_FILES_PATH)/core
```

Then I think this won't be a breaking change.

jamuir: Unfortunately, `my-dir` isn't defined when `config.mk` is processed (I think `my-dir` is intended more for use inside `Android.mk` files).  However, those variable definitions could be cleaned up.

I thought about structuring the changes here so that they don't break existing builds, but I'm not certain if that is desirable.  My thinking was that we want people to transition away from using the path `vendor/google/build`.

If my changes are pulled in, but the project is kept under the path `vendor/google/build`, then the build will fail immediately with a message that asks for the path be updated in the manifest.

jamuir:
@Dudulle:  I just discovered your PR from last year:  #41

I seem to have repeated some of the changes that you proposed.  Sorry that I did not notice it before.

jamuir:
@Blystad:  when you reviewed #41 you raised a concern about a check inside the "build" project related to the use of overlays:  https://github.com/opengapps/aosp_build/pull/41#issuecomment-243707327

The check in done inside `build/core/tasks/vendor_module_check.mk`.

Is this still a concern?

I think even overlays under "vendor/google" will be caught by that check.  But note that you can disable the check by setting `PRODUCT_RESTRICT_VENDOR_FILES:=false`.

jamuir:
@Blystad:  any update here?

Blystad:
I haven't had time to test it myself, but I trust you've done enough testing, so I'll just merge it, and then if issues arise then we'll just have to deal with them.

jamuir:
Thanks!

Issue #106, opened 2016-11-28
copying to system/vendor not allowed when building separate vendor partition


jamuir:
We hit the following error when building a Marshmallow image for bullhead:

```
Non-symlink out/target/product/bullhead/system/vendor detected!
You cannot install files to out/target/product/bullhead/system/vendor while building a separate vendor.img!
build/core/Makefile:1030: recipe for target 'out/target/product/bullhead/obj/PACKAGING/systemimage_intermediates/system.img' failed
make: *** [out/target/product/bullhead/obj/PACKAGING/systemimage_intermediates/system.img] Error 1
```

It is caused by the `PRODUCT_COPY_FILES` statements generated for "pittpatt" in `vendor/google/build/opengapps-files.mk`:

```
# Pico and higher
ifneq ($(filter $(TARGET_GAPPS_VARIANT),pico),)
# vendor/pittpatt seems to be removed on N+ (so only copy it to older than N)
ifeq ($(filter $(call get-allowed-api-levels),24),)
  PRODUCT_COPY_FILES += $(call gapps-copy-to-system,all,vendor/pittpatt)
endif
  PRODUCT_COPY_FILES += $(call gapps-copy-to-system,all,usr/srec)
endif
```

Since we build a vendor image in our build, the pittpatt files should be copied to "/vendor" rather than "/system/vendor".

jamuir:
We have fixed this locally by skipping the pittpatt copy statements when the device is bullhead (those files actually get copied to the vendor image by a statement in a separate part of our in-house build).

But rather than using device-type check, what we should really be doing is checking whether a vendor image is being built (and altering generating different `PRODUCT_COPY_FILE` statements).

Inside the "build" project, the statement "ifdef BOARD_VENDORIMAGE_FILE_SYSTEM_TYPE" is used to do that, but that doesn't work inside `vendor/google/build/opengapps-files.mk` because of the way the opengapps makefiles are included in the build system; they are processed before `device/lge/bullhead/BoardConfig.mk`.

Blystad:
I tried fixing this with https://github.com/opengapps/aosp_build/pull/65 but that didn't work, because of how stuff is processed.

See original issue https://github.com/opengapps/aosp_build/issues/63

The project I build doesn't have a vendor/ partition, so it's a bit difficult for me to test. I would welcome PRs that fixes this issue, since this seems to be a recurring issue.

[...]

PR #132, opened 2017-04-18
Update pittpatt copy statements when building a vendor image

jamuir:
The aosp build system does not allow you to copy to system/vendor when
a vendor image is being built.  Check if a vendor image is being built
and update the product-copy-files statements for pittpatt.  This fixes
issue #106.

### DIFF
--- a/opengapps-files.mk
+++ b/opengapps-files.mk
@@ -2,11 +2,22 @@
 PRODUCT_COPY_FILES += $(call gapps-copy-to-system,all,etc)
 PRODUCT_COPY_FILES += $(call gapps-copy-to-system,all,framework)
 
+# check if we are building a vendor image
+ifneq ($(CALLED_FROM_SETUP),true)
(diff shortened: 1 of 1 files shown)


opengapps-files.mk
```diff
@@ -2,11 +2,22 @@
 PRODUCT_COPY_FILES += $(call gapps-copy-to-system,all,etc)
 PRODUCT_COPY_FILES += $(call gapps-copy-to-system,all,framework)
 
+# check if we are building a vendor image
+ifneq ($(CALLED_FROM_SETUP),true)
+BUILD_VENDORIMAGE := $(shell CALLED_FROM_SETUP=true BUILD_SYSTEM=build/core \
+      command make --no-print-directory -f build/core/config.mk dumpvar-BOARD_VENDORIMAGE_FILE_SYSTEM_TYPE)
+endif
+
```
jamuir: Note that command to get the value of `BOARD_VENDORIMAGE_FILE_SYSTEM_TYPE`
is taken from the definition of the function `get_build_var` in `build/envsetup.sh`.

PR #138, opened 2017-06-14
Clarify editing the device.mk file based on my experiences

ehcloninger:
I added a couple changes to  README.md to be more precise with where the manifest exists and how to add more than one component to a build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,14 @@ PRODUCT_PACKAGES += Chrome
 
 This uses the module name. You can find the module name for a package by checking `vendor/google/build/modules/` and look at the `LOCAL_MODULE` value.
 
+If you need to include more than one module in `PRODUCT_PACKAGES`, separate the names with a space. You may also put them on separate lines, with each line terminated by a back slash. For example:
+
+```
+PRODUCT_PACKAGES += Chrome \
+    Maps \
+    GoogleEarth
+```
+
 ### Excluding packages
 You can exclude certain packages from the list of packages associated with your selected OpenGapps variant. E.g. if you have `GAPPS_VARIANT := stock`
 and want all those apps installed except for `Hangouts`, then in your `device/manufacturer/product/device.mk` just add:


README.md
```diff
@@ -7,7 +7,7 @@
 ## Getting started
 **1. Add the build system, and the wanted sources to your manifest.**
 
-Find your manifest file. Check `${ANDROID_BUILD_TOP}/.repo/manifest/`
+Find your manifest file. Check `${ANDROID_BUILD_TOP}/.repo/manifest/`. There may be a linked document in `${ANDROID_BUILD_TOP}/.repo/manifest.xml` that points to another location, such as `${ANDROID_BUILD_TOP}/.repo/manifests/default.xml`. The file you should edit should not be blank and should have several hundred lines.
```
Dudulle: We should probably recommend using local_manifests instead.
I think this has already been discussed but wasn't done for some reason.

jamuir: Different trees use different manifest configurations.  E.g. stock aosp has one giant manifest xml file (`.repo/manifests/default.xml`); Lineage uses a combination of `.repo/manifests/default.xml` and `.repo/local_manifests` (for device-specific projects that gets populated when you run `breakfast`); CodeAurora uses `.repo/manifests/${CAFTAG}.xml`.

Giving one set of instructions for everyone doesn't seem appropriate.  However, I do like the idea of giving an example of how to update the manifest in a stock AOSP tree (this is what the PR essentially does and would just need a minor edit to present it as an example).

Note that we used to use a combination of `.repo/manifests` and `.repo/local_manifests` for our in-house development, but switched to a single manifest project.  That has worked much better for us (devs no longer keep forgetting to update their local_manifests project).

ehcloninger:
This was on a cm11 branch for the Motorola XOOM, so yes, it's a bit dated. I was a bit confused in editing the repo manifest, having only worked on AOSP before. I don't mind modifying to be more generic but I'm not working in a Lineage tree at the moment. Do with the PRs as you think correct.

PR #147, opened 2017-10-01
make round icons opt-in

alviteri:
do this for android oreo launcher which can use adaptive icons

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,12 @@ If you want use PixelLauncher overriding GoogleHome you need:
 GAPPS_FORCE_PIXEL_LAUNCHER := true
 ```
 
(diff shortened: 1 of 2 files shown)


Blystad:
Why should it be opt in? Isn't round icons the new cool thing in Oreo?

alviteri:
arne pixel launcher in oreo can use adaptive icons

alviteri:
also round icons was introduced on nougat 25 api level

alviteri:
Look the adaptive icons for file manager or SIM toolkit etc etc
screenshot_20171002-020042

alviteri:
since I think this will not be merged I am closing

PR #149, opened 2017-10-09
Initial support for API 26 (Android 8.0.0)

argraur:
Added duplicate of api26hack as seen in opengapps repo and added AndroidPlatformServices


opengapps-packages.mk
```diff
@@ -34,6 +33,18 @@ GAPPS_PRODUCT_PACKAGES += \
     Turbo
 endif
 
+# api26hack
+ifneq ($(filter $(call get-allowed-api-levels),26),)
```
jamuir: this `filter` statement is consistent with the other uses inside `opengapps-packages.mk`, but after looking at all of them, I think we have the parameters swapped.

the gnu make docs say `$(filter pattern ...,text )`.  Multiple patterns can be given.

Here, `$(filter 26,$(call get-allowed-api-levels))` makes more sense to me.  Note that `get-allowed-api-levels` returns a list of integers 1..n.

This type of refactor could be done in a separate PR.

jamuir: And just to be clear, that separate PR could be done by someone else (e.g. me or anyone else who feels inclined) :-)

alviteri: @jamuir why not use GAPPS_LOCAL_OVERRIDES_PACKAGES := GoogleLoginService ?

jamuir: @alviteri :  I think I understand what you are suggesting.

Add `GAPPS_LOCAL_OVERRIDES_PACKAGES := GoogleLoginService` to the build module for `AndroidPlatformServices`; then the only change needed in `opengapps-packages.mk` is something like this:
```make
if (api level >= 26)
if (architecture is arm64)
  GAPPS_PRODUCT_PACKAGES += \
    AndroidPlatformServices
endif
endif
```

This seems a bit simpler.  I don't know of a good reason not to do it that way, but it's not my PR :-)

Blystad: I agree with using `GAPPS_LOCAL_OVERRIDE_PACKAGES` and (in a follow up commit) fix the filtering.

Btw @alviteri and @jamuir - Thanks for your numerous pull requests. I've added you as collaborators with write access, should you choose to accept it :)

argraur:
I have done everything that you suggested @jamuir @alviteri

jamuir:
thanks @argraur.

Looks like you dropped commas in the lines updated with the `filter` changes.

e.g.

```make
ifneq ($(filter 23,$(call get-allowed-api-levels)))
```
should be

```make
ifneq ($(filter 23,$(call get-allowed-api-levels)),)
```

Blystad:
Did you verify that this results in the correct behaviour? I'm a bit uncertain when I read the ifeq/ifneq without the commas, if that actually works.

opengapps-packages.mk
```diff
@@ -95,17 +103,17 @@ GAPPS_PRODUCT_PACKAGES += \
...
-ifneq ($(filter $(call get-allowed-api-levels),24),)
+ifneq ($(filter 24,$(call get-allowed-api-levels)))
 GAPPS_PRODUCT_PACKAGES += \
     GooglePrintRecommendationService \
     GoogleExtServices \
     GoogleExtShared
 endif
 
-ifneq ($(filter $(TARGET_GAPPS_VARIANT),super),)
+ifneq ($(filter stock,$(TARGET_GAPPS_VARIANT)))
```
jamuir: shouldn't this be `super`?

argraur:
Done @jamuir

opengapps-packages.mk
alviteri: thank you @Blystad

alviteri:
reopened here https://github.com/opengapps/aosp_build/pull/151 @jamuir @argraur @Blystad

jamuir:
@alviteri :  we should pick up argraur's other changes (i.e. the refactoring of the "filter" statements) in another PR.

alviteri:
@jamuir I have picked and have error in build ... Related to webview overlays .... Are you sure that refactor works ?

jamuir:
no, but it should work :-)

I will give it a try in my local tree to see if I can reproduce the error.

jamuir:
@argraur :  I tried your branch and hit the error below.  It seems there is still a comma missing.  It is probably best to do refactoring like that in a separate PR.

Also, it turns out that the variable `TARGET_ARCH` is not defined when `opengapps-packages.mk` is processed, so the existing architecture check we had there did not work properly (and same for the new one you added).  alviteri is fixing that in #151.

```
$ make installclean
============================================
PLATFORM_VERSION_CODENAME=REL
PLATFORM_VERSION=7.1.2
TARGET_PRODUCT=aosp_bullhead
TARGET_BUILD_VARIANT=userdebug
TARGET_BUILD_TYPE=release
TARGET_BUILD_APPS=
TARGET_ARCH=arm64
TARGET_ARCH_VARIANT=armv8-a
TARGET_CPU_VARIANT=cortex-a53
TARGET_2ND_ARCH=arm
TARGET_2ND_ARCH_VARIANT=armv7-a-neon
TARGET_2ND_CPU_VARIANT=cortex-a53.a57
HOST_ARCH=x86_64
HOST_2ND_ARCH=x86
HOST_OS=linux
HOST_OS_EXTRA=Linux-4.4.0-96-generic-x86_64-with-Ubuntu-16.04-xenial
HOST_CROSS_OS=windows
HOST_CROSS_ARCH=x86
HOST_CROSS_2ND_ARCH=x86_64
HOST_BUILD_TYPE=release
BUILD_ID=NZH54D
OUT_DIR=out
============================================
Running kati to generate build-aosp_bullhead-installclean.ninja...
out/build-aosp_bullhead-installclean.ninja is missing, regenerating...
vendor/opengapps/build/opengapps-packages.mk:154: *** extraneous `endif'.
build/core/ninja.mk:163: recipe for target 'out/build-aosp_bullhead-installclean.ninja' failed
make: *** [out/build-aosp_bullhead-installclean.ninja] Error 1

#### make failed to build some targets (2 seconds) ####
```

alviteri:
@jamuir when I cherry-picked this and I got build working with the following error

```
[ 75% 56723/75068] AAPT2 link /mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/package-res.apk
FAILED: /mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/package-res.apk
/bin/bash /mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/package-res.apk.rsp
/mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/flat-res/vendor/opengapps/build/overlay/webview/21/frameworks/base/core/res/res/values_config.arsc.flat: error: resource string/config_webViewPackageName does not override an existing resource.
/mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/flat-res/vendor/opengapps/build/overlay/webview/21/frameworks/base/core/res/res/values_config.arsc.flat: note: define an tag or use --auto-add-overlay.
error: failed parsing overlays.
```

so I think 
```
https://github.com/opengapps/aosp_build/pull/149/commits/1be10d0ecf6ee872e1f2035903bd85161185a770#diff-6f8f370c5563e7ea06c5741d8703bb51R146
```
is not working

jamuir:
@argraur :  thanks for opening your PR.  Sorry that it did not get merged, but it did prompt us to fix an important pre-existing problem.  Hopefully, you can contribute another PR in the future.

I created a new PR with changes to the "filter" statements ( #154 ) so we can test if the error alviteri hit comes back.

PR #151, opened 2017-10-12
Master oreo

alviteri:


alviteri:
[ 75% 56723/75068] AAPT2 link /mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/package-res.apk
FAILED: /mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/package-res.apk 
/bin/bash /mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/package-res.apk.rsp
/mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/flat-res/vendor/opengapps/build/overlay/webview/21/frameworks/base/core/res/res/values_config.arsc.flat: error: resource string/config_webViewPackageName does not override an existing resource.
/mnt/users/david/build/target/product/poplar/obj/APPS/framework-res_intermediates/flat-res/vendor/opengapps/build/overlay/webview/21/frameworks/base/core/res/res/values_config.arsc.flat: note: define an <add-resource> tag or use --auto-add-overlay.
error: failed parsing overlays.

opengapps-packages.mk
```diff
@@ -34,6 +33,16 @@ GAPPS_PRODUCT_PACKAGES += \
     Turbo
 endif
 
+ifneq ($(filter $(call get-allowed-api-levels),26),)
```
alviteri: @jamuir doing this because GoogleLoginService is a PRIVILEGED app and can't be override or it seems ... going to build right now

alviteri:
since the app is available only on arm64 repo there is not needed to add a conditional for specific arch and can work like this ... @jamuir @Blystad

Blystad:
Won't it then print an error since LOCAL_SRC_FILES would be empty? (since it can't find the APK)

alviteri:
generally on aosp when a package is not found and defined build_system ignores it ... is not the same here ?

Blystad:
Correct (and I hate that behaviour, but that's a discussion for another time), but, you define the module here, but with no source files. I did a simple test locally, by trying to compile the following `Android.mk` using `mm`:

```
LOCAL_PATH := .
include $(CLEAR_VARS)
include $(GAPPS_CLEAR_VARS)
LOCAL_MODULE := TestModule
LOCAL_PACKAGE_NAME := some.none.existing.module
LOCAL_PRIVILEGED_MODULE := true

```

and I got the error:

```
PRODUCT_COPY_FILES device/ti/common-open/audio/audio_effects.conf:system/etc/audio_policy.conf ignored.
No private recovery resources for TARGET_DEVICE x7k
make: Entering directory `/home/blystad/x7k/android/lollipop'
target Prebuilt: TestModule (/home/blystad/x7k/android/lollipop/out/target/product/x7k/obj/APPS/TestModule_intermediates/package.apk)
acp: omitting directory '.'
mv: cannot stat `/home/blystad/x7k/android/lollipop/out/target/product/x7k/obj/APPS/TestModule_intermediates/package.apk': No such file or directory
make: *** [/home/blystad/x7k/android/lollipop/out/target/product/x7k/obj/APPS/TestModule_intermediates/package.apk] Error 1
make: Leaving directory `/home/blystad/x7k/android/lollipop'
```

You could add a check in the Android.mk for TARGET_ARCH (so that the file would be "empty", if you catch my drift, if the TARGET_ARCH is not arm64). _then_ I think it would work fine.

alviteri:
@Blystad I have a problem
adding ifneq ($(filter arm64, $(TARGET_ARCH)),) doesn't work .. it works in Android.mk as you suggested 

same for Facelock ... can we change those ?

jamuir:
confirmed here.

in my tree, I include `opengapps-packages.mk` at the bottom of `device/lge/bullhead/device.mk` (which is what we recommend in the README.md file).

When `opengapps-packages.mk` is processed, `TARGET_ARCH` isn't defined yet (that variable gets defined in BoardConfig.mk, which is processed later).

I'm not sure of the best way to fix this.

The architecture check could be moved into the Android.mk file; or we could invoke "make" to obtain the value of `TARGET_ARCH` inside `opengapps-packages.mk` (see b8c6c5d95 for an example).

alviteri:
The best way is to define in Android.mk as Arne said

alviteri:
It worked here

jamuir:
That sounds good.

I guess it will look something like this:

```makefile
ifneq ($(filter arm64, $(TARGET_ARCH)),) 

LOCAL_PATH := .
include $(CLEAR_VARS)
include $(GAPPS_CLEAR_VARS)
LOCAL_MODULE := TestModule
LOCAL_PACKAGE_NAME := some.none.existing.module
LOCAL_PRIVILEGED_MODULE := true

endif
```

We use that pattern in some of our vendor projects.

alviteri:
Yeah I did it and worked

alviteri:
can this be merged then ?

jamuir:
done (using my new merge powers) :-)

Blystad:
:-)  Feel free to use them

PR #153, opened 2017-10-13
fix TARGET_ARCH cond

alviteri:
this is called early and don't take effect in build.....
fix this moving the cond to Android.mk


jamuir:
looks good to me.

alviteri:
ping @Blystad

PR #160, opened 2017-12-08
build: fix up for python3

invisiblek:
* Use print()
* Python-ize the lookup of files in opengapps_sources

Change-Id: I5eca400d356f879e707fa5b70f4a72beb5f0df65


jamuir:
looks good.

it will still work with python2, right?

invisiblek:
Yea, in fact python3 doesn't build without some major changes to the build system (which I'm working on yet), but I tested a py2 build and things seem fine.

jamuir:
great.  thanks for confirming.

PR #165, opened 2018-02-06
Move libfrsdk.so from vendor to system

Myself5:
On most devices we don't build vendor images and use the Stock vendor images.

On stock, the lib is on system (Verified on the Pixel 2 XL running 8.1), so move it to system here too.


jamuir:
Thanks for your PR.

I see that the lib also resides under system in the Pixel (marlin) 8.0 factory image.

Issue #134, opened 2017-05-06
Phone recognized as Pixel in Google Photos


vasilmkd:
Nexus 6 build with stock OpenGapps package. There are no build.prop tweaks, yet the phone is recognized as a Pixel in Google Photos, which doesn't let me change the backup type. My Google Drive storage gets depleted every time I backup a photo.

It is a user build, so I have no root. Any suggestions?

mfonville:
Ah, I recently changed the file `/etc/sysconfig/nexus.xml` to be again as it is on Pixel devices. It is just important that this file should **only** be installed on Pixel devices and not on ROMs for other devices. Aosp_build will need an update to take this into account.

vasilmkd:
But if I manually delete the two strings before my build, I should be in the clear right?

mfonville:
yes, but even better would be to completely drop the file if it ain't a pixel, to avoid issues with setupwizard too (since android 7.1)

vasilmkd:
Right, because the setup wizard is also Pixel-esque! Thank you very much!

mfonville:
no problem, but not sure if closing the issue is the right thing; the aosp_build script needs an update for this so probably @Blystad wants to have a look at this

Blystad:
Right now we just blindly take everything from all/etc. I guess we only want "google.xml" by default. Then Nexus users can add their own PRODUCT_COPY_FILES for the other files, if they so desire.

PR #166, opened 2018-02-19
Stop including Nexus/Pixel sysconfig xmls on all devices

Blystad:
This fixes issues where any device building with OpenGapps would
believe it's a Google Pixel device.

Fixes issue #134

### DIFF
--- a/modules/GoogleDialer/Android.mk
+++ b/modules/GoogleDialer/Android.mk
@@ -9,3 +9,15 @@ GAPPS_LOCAL_OVERRIDES_MIN_VARIANT :=
 GAPPS_LOCAL_OVERRIDES_PACKAGES := Dialer
 
 include $(BUILD_GAPPS_PREBUILT_APK)
+
+include $(CLEAR_VARS)
+include $(GAPPS_CLEAR_VARS)
+
+LOCAL_MODULE := GoogleDialer_dialer_experience.xml
+LOCAL_MODULE_STEM := dialer_experience.xml
+LOCAL_MODULE_TAGS := optional
+LOCAL_MODULE_CLASS := ETC
+LOCAL_SRC_FILES := $(GAPPS_SOURCES_PATH)/all/etc/sysconfig/dialer_experience.xml
+LOCAL_MODULE_PATH := $(TARGET_OUT_ETC)/sysconfig
+
+include $(BUILD_PREBUILT)
--- a/opengapps-files.mk
+++ b/opengapps-files.mk
@@ -1,6 +1,46 @@
 # Always needed
-PRODUCT_COPY_FILES += $(call gapps-copy-to-system,all,etc)
 PRODUCT_COPY_FILES += $(call gapps-copy-to-system,all,framework)
+
+GAPPS_NEXUS_CODENAMES += \
+    full_maguro \
+    full_toro \
+    full_toroplus \
+    aosp_grouper \
+    aosp_tilapia \
+    aosp_manta \
+    aosp_mako \
+    aosp_flo \
+    aosp_deb \
+    aosp_hammerhead \
+    aosp_flounder \
+    aosp_shamu \
+    aosp_angler \
+    aosp_bullhead
+
+GAPPS_PIXEL_CODENAMES += \
+    aosp_marlin \
+    aosp_sailfish \
+    aosp_muskie \
+    aosp_taimen \
+    aosp_wahoo
+
+gapps_etc_files := $(call gapps-copy-to-system,all,etc)
+
+# Remove google_build.xml on non-Pixel devices
+ifeq ($(filter $(GAPPS_PIXEL_CODENAMES),$(TARGET_PRODUCT)),)
+  gapps_etc_files := $(filter-out %sysconfig/google_build.xml,$(gapps_etc_files))
+endif
+
+# Remove nexus.xml on non-Nexus devices
+ifeq ($(filter $(GAPPS_NEXUS_CODENAMES),$(TARGET_PRODUCT)),)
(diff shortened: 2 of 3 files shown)


opengapps-files.mk
```diff
@@ -1,7 +1,47 @@
...
+    manta \
+    mako \
+    flo \
+    deb \
+    hammerhead \
+    flounder \
+    shamu \
+    angler
+
+GAPPS_PIXEL_CODENAMES += \
```
jamuir: do we need "sailfish" here, too?

modules/GoogleDialer/Android.mk
```diff
@@ -9,3 +9,14 @@ GAPPS_LOCAL_OVERRIDES_MIN_VARIANT :=
 GAPPS_LOCAL_OVERRIDES_PACKAGES := Dialer
 
 include $(BUILD_GAPPS_PREBUILT_APK)
```
jamuir: minor suggestion:  maybe add a blank line or a comment before CLEAR_VARS so it is more obvious that there are actually two modules defined in this file.

modules/GoogleDialer/Android.mk
```diff
@@ -9,3 +9,14 @@ GAPPS_LOCAL_OVERRIDES_MIN_VARIANT :=
...
+include $(GAPPS_CLEAR_VARS)
+
+LOCAL_MODULE := GoogleDialer_dialer_experience.xml
+LOCAL_MODULE_STEM := dialer_experience.xml
+LOCAL_MODULE_TAGS := optional
+LOCAL_MODULE_CLASS := ETC
+LOCAL_SRC_FILES := $(GAPPS_SOURCES_PATH)/all/etc/sysconfig/dialer_experience.xml
+LOCAL_MODULE_PATH := $(TARGET_OUT_ETC)/sysconfig
+
+include $(BUILD_PREBUILT)
```
jamuir: is there an advantage to having a module defined for GoogleDialer_dialer_experience.xml?

since it is just a single file that needs to be copied (or not copied), why not just use PRODUCT_COPY_FILES?

Blystad: Just that if someone runs `mmm vendor/opengapps/build/modules/GoogleDialer` the file gets included. It also makes the dependency more explicit.

opengapps-files.mk
Blystad: Yes. And I just noticed the TARGET_PRODUCT is named aosp_sailfish and aosp_muskie instead...so I'll have to handle that. Bleh.

Blystad:
Fixed product names now (checked the various projects on android.googlesource.com)

alviteri:
Looks good

modules/GoogleDialer/Android.mk
jamuir: to make the dependency really explicit, you could do this:

```diff
$ git diff
diff --git a/modules/GoogleDialer/Android.mk b/modules/GoogleDialer/Android.mk
index c49aa3a..b772773 100644
--- a/modules/GoogleDialer/Android.mk
+++ b/modules/GoogleDialer/Android.mk
@@ -8,6 +8,9 @@ LOCAL_PRIVILEGED_MODULE := true
 GAPPS_LOCAL_OVERRIDES_MIN_VARIANT :=
 GAPPS_LOCAL_OVERRIDES_PACKAGES := Dialer
 
+# express dependency on xml file
+GoogleDialer: GoogleDialer_dialer_experience.xml
+
 include $(BUILD_GAPPS_PREBUILT_APK)
 
 include $(CLEAR_VARS)
```
This way the xml file will get copied even when someone runs `m GoogleDialer`.

jamuir:
I made another minor suggestion above, but it can be ignored :-)

Looks good to me, too.

Merge whenever you are ready.

modules/GoogleDialer/Android.mk
jamuir: after looking at some of the other makefiles, I see you can do the same thing using `LOCAL_REQUIRED_MODULES`.  That is probably the preferred way.

```diff
$ git diff
diff --git a/modules/GoogleDialer/Android.mk b/modules/GoogleDialer/Android.mk
index c49aa3a..d75a84c 100644
--- a/modules/GoogleDialer/Android.mk
+++ b/modules/GoogleDialer/Android.mk
@@ -8,6 +8,8 @@ LOCAL_PRIVILEGED_MODULE := true
 GAPPS_LOCAL_OVERRIDES_MIN_VARIANT :=
 GAPPS_LOCAL_OVERRIDES_PACKAGES := Dialer
 
+LOCAL_REQUIRED_MODULES := GoogleDialer_dialer_experience.xml
+
 include $(BUILD_GAPPS_PREBUILT_APK)
 
 include $(CLEAR_VARS)
```

jamuir:
@Blystad :  is it ok with you if I add the following line to `GoogleDialer/Android.mk`?

```make
LOCAL_REQUIRED_MODULES := GoogleDialer_dialer_experience.xml
```

alviteri:
@jamuir +1 since aosp dialer doesn't need it ... Let it included only for Dialer flag

alviteri:
GAPPS_FORCE_DIALER_OVERRIDES := true include it ....

Blystad:
James, sure, go for it :)

On 19 Mar 2018 6:33 pm, "James Muir" <notifications@github.com> wrote:

@Blystad <https://github.com/blystad> : is it ok with you if I add the
following line to GoogleDialer/Android.mk?

LOCAL_REQUIRED_MODULES := GoogleDialer_dialer_experience.xml

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/opengapps/aosp_build/pull/166#issuecomment-374298694>,
or mute the thread

PR #173, opened 2018-03-28
Include BUILD_PREBUILT only when LOCAL_SRC_FILES available

### LINKED ISSUE #172: build failure on x86_64, cp: missing destination file operand after '.../GoogleCamera_intermediates/package.apk'
Full error messages:
```
[  0% 1/16329] target Prebuilt: GoogleCamera (out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk)
FAILED: out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk
/bin/bash -c "(rm -f out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk ) && (cp  out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk ) && (if (zipinfo out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk 'lib/*.so' 2>/dev/null | grep -v ' stor ' >/dev/null) ; then rm -rf out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/uncompressedlibs && mkdir out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/uncompressedlibs; unzip out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk 'lib/*.so' -d out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/uncompressedlibs && zip -d out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk 'lib/*.so' && ( cd out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/uncompressedlibs && find lib -type f | sort | zip -D -X -0 ../package.apk -@ ) && rm -rf out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/uncompressedlibs; fi ) && (if ! out/host/linux-x86/bin/zipalign -c -p 4 out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk >/dev/null ; then mv out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk.unaligned; out/host/linux-x86/bin/zipalign -f -p 4 out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk.unaligned out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk.aligned; mv out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk.aligned out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk; fi )"
cp: missing destination file operand after 'out/target/product/generic_x86_64/obj/APPS/GoogleCamera_intermediates/package.apk'
Try 'cp --help' for more information.
[  0% 6/16329] target Prebuilt: GoogleTTS (out/target/product/generic_x86_64/obj/APPS/GoogleTTS_intermediates/package.apk)
Archive:  out/target/product/generic_x86_64/obj/APPS/GoogleTTS_intermediates/package.apk
  inflating: out/target/product/generic_x86_64/obj/APPS/GoogleTTS_intermediates/uncompressedlibs/lib/x86/libtts_android.so
deleting: lib/x86/libtts_android.so
  adding: lib/x86/libtts_android.so (stored 0%)
[  0% 8/16329] Copy: out/target/common/obj/JAVA_LIBRARIES/core-all_intermediates/classes.jar
ninja: build stopped: subcommand failed.
07:26:45 ninja failed with: exit status 1
make: *** [run_soong_ui] Error 1
Command exited with non-zero status 2
```

From [core/prebuilt_apk.mk](https://github.com/opengapps/aosp_build/blob/master/core/prebuilt_apk.mk), included from [modules/GoogleCamera/Android.mk](https://github.com/opengapps/aosp_build/blob/master/modules/GoogleCamera/Android.mk), it tries to locate appropriate prebuilt apk files for **all**, **$(TARGET_ARCH)** and **$(TARGET_2ND_ARCH)**. However, while GoogleCamera is currently only available under **arm** and **arm64**, an empty `LOCAL_SRC_FILES` is returned, which is then referenced in AOSP [build/core/prebuilt_internal.mk](https://android.googlesource.com/platform/build/+/android-8.1.0_r1/core/prebuilt_internal.mk#379) as `my_prebuilt_src_file`, which is then used as first file operand for command `cp`.

To avoid this error, I'd suggest to move that `include $(BUILD_PREBUILT)` line into the following `ifdef LOCAL_SRC_FILES`-braced block in [core/prebuilt_apk.mk](https://github.com/opengapps/aosp_build/blob/master/core/prebuilt_apk.mk), line 52.

vicamo:
Fixes #172.

Signed-off-by: You-Sheng Yang <vicamo@gmail.com>

### DIFF
--- a/core/prebuilt_apk.mk
+++ b/core/prebuilt_apk.mk
@@ -46,10 +46,10 @@ else
   endif
 endif
 
-include $(BUILD_PREBUILT)
-
 # generate mk file of shared library for kitkat
 ifdef LOCAL_SRC_FILES
+  include $(BUILD_PREBUILT)
+
   ifeq ($(filter 21,$(call get-allowed-api-levels)),)
     ifneq ($(call find-libs-in-apk,$(TARGET_ARCH),$(LOCAL_SRC_FILES)),)
       $(shell unzip -qqq -j -o "$(LOCAL_SRC_FILES)" "$(call get-lib-search-path, $(TARGET_ARCH))" -x lib/*/crazy/* -d "$(GAPPS_SOURCES_PATH)"/$(TARGET_ARCH)/lib/19/lib_from_app 2>/dev/null)


jamuir:
@vicamo :  thanks for your PR!  sorry that we did not get to it sooner.

I am having a difficult time deciding what the right behaviour is here.

The build functions in `aosp_build` attempt to automatically find the right source file (apk) based on the target architecture.  If they cannot find any apk, then that resulted in a (non-obvious) build error.  But now, with your change, if they cannot find any apk, then they will just silently skip that module.

You could argue that an error is appropriate in this case (and we could use `$(error ...)` to emit a clear error message).  Then, to fix x86 builds, we would need to add logic to exclude the Camera module.

I'm not sure what is right (skip or error).

@Blystad @alviteri :  any opinion on this?

If we are going to skip over a requested module, then we at least need to emit a clear warning about that.  Something like this:

```make
ifndef LOCAL_SRC_FILES
  $(warning Unable to locate source files for module $(LOCAL_MODULE))
else
  include $(BUILD_PREBUILT)
  # generate mk file of shared library for kitkat
  ifeq ($(filter 21,$(call get-allowed-api-levels)),)
...
```

jamuir:
Another way to fix the x86 build would have been to use `GAPPS_EXCLUDED_PACKAGES` in your device.mk file:

```make
GAPPS_EXCLUDED_PACKAGES += GoogleCamera
```

vicamo:
Hi, just thought it would be nice for general users to pass the build with only default options. You might add more apps in the future that would either fail a build or require additional modification to existing `device.mk`.

If you'd like to know what's missing for the arch, I'd suggest adding another flag in `buildspec.mk` instead.

jamuir:
> If you'd like to know what's missing for the arch, I'd suggest adding another flag in buildspec.mk instead.

I know developers can create a `buildspec.mk` file in the root of their tree to set build defaults, but I'm not sure what you are suggesting.  Can you tell me more?

alviteri:
@jamuir I prefer the error to advertise the user ...

vicamo:
@jamuir Sorry, was outside all day. I mean something like:
```Makefile
ifeq ($(strip $(LOCAL_SRC_FILES)),)
ifeq ($(strip $(GAPPS_WARN_UNAVAILABLE_PACKAGES)),true)
$(error Package $(LOCAL_MODULE) unavailable under current arch $(TARGET_ARCH)$(if $(TARGET_2ND_ARCH),/$(TARGET_2ND_ARCH)).)
endif
else
  include $(BUILD_PREBUILT)
  ...
endif
```
So for those who are interested in missed apps, you may add **GAPPS_WARN_UNAVAILABLE_PACKAGES := true** to your `buildspec.mk`.

In my opinion, upstream projects should try hard to make it easy for downstream users while preserving flexibility as possible. Firstly, knowing "GoogleCamera is unavailable on x86_64" doesn't change anything in down stream projects. All that I can do is to remove the line that includes GoogleCamera, or, if I'd like to spent some more time, find out why it's trying to include an unavailable app under current configuration and file a bug, and/or a PR. Even then I still cannot help the ultimate problem "GoogleCamera is unavailable on x86_64". So it's a blocker for all others while there is no viable solution except work-around like this PR.

Secondly, for every new app you add, you're doing great favor to down stream projects and that's very appreciated. But if that breaks any currently successful builds and/or requires modification to all those builds, you're bringing goodies to some and inconveniences to others. For those architectures that GoogleCamera is unavailable, and before you include GoogleCamera in other architectures, they have been, somehow, addressed this issue and are living with happiness. After you include GoogleCamera in other architectures, which is really irrelevant to them, however now their builds failed due to #172. This doesn't make sense to me. It doesn't have to fail everybody else to complete itself. It could have been done better.

Thirdly, you can still have as much methods to know what's missing under current config as you want without blocking builds based on x86/x86_64. Something like the first paragraph would do.

That's my 2 cents.

Blystad:
I'd rather have an error by default, than something you have to explicitly opt-in to.

Instead of that obscure `cp` error we could do something like:
```
ifeq ($(strip $(LOCAL_SRC_FILES)),)
$(error Package $(LOCAL_MODULE) unavailable under current arch $(TARGET_ARCH)$(if $(TARGET_2ND_ARCH),/$(TARGET_2ND_ARCH)).)
endif
```

and then for apps that don't exist for a given architecture, like GoogleCamera, that app can have a check of the TARGET_ARCH and TARGET_2ND_ARCH, and if it's x86 then just don't build it.

Maybe add a variable like

```
GAPPS_APP_ARCH := arm arm64
```

to GoogleCamera, and then based on this variable, if the TARGET_ARCH/TARGET_2ND_ARCH doesn't match - skip it. If it does match, and the file is not found - show error.

vicamo:
That **GAPPS_APP_ARCH** simply duplicates upper half of this file.

By the way, where do these apks come from? Didn't find documents for this.

Blystad:
https://github.com/opengapps/apkcrawler

jamuir:
@vicamo :  I know that is not what you prefer, but would you consider updating your PR to emit an informative error message (like the examples above) when LOCAL_SRC_FILES is not defined?  The error message could even suggest a work-around to the user (e.g. using `GAPPS_EXCLUDED_PACKAGES += $(LOCAL_MODULE)`).  If you made that change, then we could merge it right away.

vicamo:
Sorry. It doesn't have to be merged as it is, but I don't want to leave my name on a commit that I don't agree. Thanks for your time.

jamuir:
ok.  fair enough :-)

I hope you will consider contributing to the project in the future even though we did not agree this time.  Thanks!

PR #178, opened 2018-04-19
Use GAPPS_EXCLUDED_PACKAGES to guard invalid module error

jamuir:
For modules that use "include $(BUILD_GAPPS_PREBUILT_APK)", if we
cannot locate an architecture compatible apk to install, then we treat
that module as invalid and issue an error.  To work around such
errors, the user might use GAPPS_EXCLUDED_PACKAGES, so check that
variable first before erroring-out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ You can add packages from versions higher then your set version. E.g. if you wan
 In your `device/manufacturer/product/device.mk` just add, for example:
 
 ```makefile
-PRODUCT_PACKAGES += Chrome
+GAPPS_PRODUCT_PACKAGES += Chrome
 ```
 
 This uses the module name. You can find the module name for a package by checking `vendor/opengapps/build/modules/` and look at the `LOCAL_MODULE` value.
--- a/core/prebuilt_apk.mk
+++ b/core/prebuilt_apk.mk
@@ -47,7 +47,13 @@ else
 endif
 
 ifndef LOCAL_SRC_FILES
-  $(error Unable to find APK for $(LOCAL_MODULE))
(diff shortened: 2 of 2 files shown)


jamuir:
@Blystad :  you had mentioned adding `$(strip ...)` someplace but I'm not sure if it is necessary.

btw, here is an example of the error message:

```
vendor/opengapps/build/core/prebuilt_apk.mk:53: Unable to find an architecture compatible APK for package xxxcom.google.android.contacts defined in module xxxGoogleContacts.
vendor/opengapps/build/core/prebuilt_apk.mk:54: You can try using GAPPS_EXCLUDED_PACKAGES += xxxGoogleContacts to get past this error.
vendor/opengapps/build/core/prebuilt_apk.mk:55: *** Invalid build module: xxxGoogleContacts.
build/core/ninja.mk:163: recipe for target 'out/build-aosp_sailfish.ninja' failed
make: *** [out/build-aosp_sailfish.ninja] Error 1
```

jamuir:
One consequence of our new error checking is that errors will be emitted even for modules that a user isn't going to build.

e.g. if the user is building "pico" and there is something wrong with the module for, say, `GoogleEarth` (installed as part of "super"), then their build will fail and they will have to exclude that module.

I'm not sure if that is desirable since it will undoubtedly cause some confusion, even though I do like errors that help us uncover bad modules.

if we wanted to change this (i.e. so we only emit errors for modules that a user is going to build), then we could also check the contents of `GAPPS_PRODUCT_PACKAGES` .

jamuir:
I've added a further restriction on when the error message is emitted.

Blystad:
LGTM!

PR #180, opened 2018-04-25
Add support for "lzip --decompress"

jamuir:
If "lunzip" fails, then try "lzip --decompress".

### DIFF
--- a/core/find_apk.sh
+++ b/core/find_apk.sh
@@ -11,10 +11,17 @@ getlatestapk() {
 " # We set IFS to newline here so that filenames containing spaces are
   # handled correctly in the for-loops below.
 
+  local LZ_DECOMPRESS_CMD="false"
+  if command -v lunzip > /dev/null; then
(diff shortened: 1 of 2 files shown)


core/find_apk.sh
```diff
@@ -14,7 +14,7 @@ getlatestapk() {
   # decompress apks
   # some apks are lz compressed to work around github file-size limits.
   for foundapklz in $(find "$1" -iname '*.apk.lz'); do
-    lunzip --keep "$foundapklz"
+    lunzip --keep "$foundapklz" &> /dev/null || lzip --decompress --keep "$foundapklz"
```
Blystad: What about testing if lunzip exists (if [[ -e "$(command -v lunzip)" ]]) ? instead of sending the result to /dev/null

jamuir:
I just pushed an update.

Instead of `if [[ -e "$(command -v lunzip)" ]]`, I tried to do it in a way that does not cause a new subshell to be spawned :-)

PR #182, opened 2018-06-04
Fix multiarch support

Ganster41:
Android build system doesn't extract JNI libs from prebuilt APK files on latest Android versions, and ignoring LOCAL_PREBUILT_JNI_LIBS value. It repacks them with no compression instead, and uses directly from an APK, but for target arch only. This value will be ignored by older Android versions, so checking of target Android version isn't necessary.

E.g. if you specified arm64 as your TARGET_ARCH, but your APK has 32-bit libs only, you'll get UnsatisfiedLinkError on runtime.

So, you must build whole APK to the 2nd arch, to let VM know that APK has only this arch support.

Signed-off-by: Ganster41 <ganster0x29@gmail.com>


jamuir:
Thanks for the PR!

PR #190, opened 2018-10-19
Camera and Nexus/Pixel build fixes

Rashed97:
Since only Nexus and Pixel devices can use the experimental camera JARs, only ship on those devices. Also, since not all targets will always begin with aosp_ or full_, use a wildcard. Also add experimental2017 camera JARs and ship the correct JARs for the correct generation of devices.


Rashed97:
Corresponding patch for experimental2017 addition: https://github.com/opengapps/all/pull/31

PR #195, opened 2018-10-25
make pixel launcher default launcher for oreo and newer android releases

alviteri:

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ If you want to include Chrome on a non-full build you need:
 GAPPS_FORCE_BROWSER_OVERRIDES := true
 ```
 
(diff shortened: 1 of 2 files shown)


jamuir:
Other than re-wording the README change, this looks good to me.

We have been using GAPPS_FORCE_PIXEL_LAUNCHER:=true in our builds since Nougat :-)

alviteri:
james I don't want be invasive in README so I am merging this as is .... probably we can rework the README later ....

jamuir:
of course it can be changed later :-)  the point is that since you decided to change it now, it should be changed in a good way.  Why is "Default" capitalized?  Also, "If you want use" is not correct.

PR #197, opened 2018-10-29
Add ActionsServices

tsubus:


opengapps-packages.mk
```diff
@@ -127,6 +127,7 @@ GAPPS_PRODUCT_PACKAGES += \
     Wallet \
     DMAgent \
     CarrierServices \
+    ActionsServices \
```
jamuir: Wasn't ActionsServices added in Pie?  In that case, it should only be added to the build with api level 28 or higher.

tsubus: you are right about that, let me fix it up

tsubus:
that should do it

jamuir:
Thanks!

Looks good to me.

Merging.

PR #202, opened 2018-12-11
overlay: Define Google Assistant as default

jhenrique09:
Google introduced that on r21 patch

https://github.com/PixelExperience/frameworks_base/commit/7105e37a8c6d4dcf95695298486b25b8c43a6a7e and https://github.com/PixelExperience/frameworks_base/commit/3a1a64c045ffe89823d8cf2df17655d119f967ba


jhenrique09:
Idk if this should live on dedicated folder called "nano" (since Google app is included on that)

alviteri:
Create an overlay for api28 since this will break older android versions

alviteri:
Force push in 1 commit please then I will merge

PR #204, opened 2019-01-28
Install keyboard swipe lib for nano+ packages.

maniac103:


jamuir:
thanks for opening this.  I had noticed that GoogleKeyboard was immediately trying to download something (even when there was no network) when ever I booted a newly built image.  Maybe it was this lib.

modules/LatinImeGoogle/Android.mk
```diff
@@ -5,6 +5,11 @@ LOCAL_MODULE := LatinImeGoogle
 LOCAL_PACKAGE_NAME := com.google.android.inputmethod.latin
 
 GAPPS_LOCAL_OVERRIDES_MIN_VARIANT :=
-GAPPS_LOCAL_OVERRIDES_PACKAGES := LatinIME
+GAPPS_LOCAL_OVERRIDES_PACKAGES := LatinIME libjni_latinimegoogle
```
jamuir: I'm not sure why `libjni_latinimegoogle` is added to `GAPPS_LOCAL_OVERRIDES_PACKAGES` here.

I think it makes sense to add this instead:

`LOCAL_REQUIRED_MODULES := libjni_latinimegoogle`

After that, you can then also drop the change to `opengapps-packages.mk`.

maniac103: The intent of this PR is to make swipe input work with LatinIME (the AOSP keyboard). Thus, I enabled the lib package by default and disable it again if LatinImeGoogle is installed to avoid potential conflicts.
If you're saying LatinImeGoogle is missing this lib as well (I have no idea whether that's the case, I don't use stock package), I can drop the change in this line, but the package addition needs to stay, as otherwise my use case doesn't work anymore ;-)

(BTW, AFAICT the zip installer also always installs this lib)

maniac103: Sure ... update in what way exactly?

jamuir:
@alviteri :  this wasn't really to be merged yet.

alviteri:
+1 and thanks ... I didn't check ... was from phone

maniac103:
@jamuir LGTM :+1:

PR #205, opened 2019-01-28
Properly separate lunzip arguments.

maniac103:
At least on my Fedora 29 machine, having command and argument in one
shell variable causes them to be used as argv[0], IOW, the script will
run "lzip --decompress" --keep [...] which obviously leads to failure.


jamuir:
thanks for this fix!

PR #206, opened 2019-02-06
Add StorageManagerGoogle module

tsubus:
See title


alviteri:
See readme and why we don't include Google messenger by default ... 

https://github.com/opengapps/aosp_build/blob/master/README.md

tsubus:
yea just found it out by my own, was about to edit it. same for googlepay (wallet). storagemanager still missing

tsubus:
It might be nice to one day align the module names in aosp_build with the package comparison table names, to avoid confusion

PR #209, opened 2019-02-17
use soong_ui.bash instead of config.mk on pie and newer

Stricted:
Change-Id: I84f481cbda0cd6dfce296a542c52b9a219ddcdd4

### DIFF
--- a/opengapps-files.mk
+++ b/opengapps-files.mk
@@ -95,8 +95,12 @@ PRODUCT_COPY_FILES += $(gapps_etc_files) $(gapps_framework_files)
 
 # check if we are building a vendor image
 ifneq ($(CALLED_FROM_SETUP),true)
+ifneq ($(filter 28, $(call get-allowed-api-levels)),)
+BUILD_VENDORIMAGE := $(shell CALLED_FROM_SETUP=true build/soong/soong_ui.bash --dumpvar-mode BOARD_VENDORIMAGE_FILE_SYSTEM_TYPE)
+else
 BUILD_VENDORIMAGE := $(shell CALLED_FROM_SETUP=true BUILD_SYSTEM=build/core \
       command make --no-print-directory -f build/core/config.mk dumpvar-BOARD_VENDORIMAGE_FILE_SYSTEM_TYPE)
+endif
 endif
 
 # Pico and higher


jamuir:
thanks for your PR!

opengapps-files.mk
```diff
@@ -95,8 +95,12 @@ PRODUCT_COPY_FILES += $(gapps_etc_files) $(gapps_framework_files)
 
 # check if we are building a vendor image
 ifneq ($(CALLED_FROM_SETUP),true)
+ifneq ($(filter 26 27, $(PLATFORM_SDK_VERSION)),)
 BUILD_VENDORIMAGE := $(shell CALLED_FROM_SETUP=true BUILD_SYSTEM=build/core \
       command make --no-print-directory -f build/core/config.mk dumpvar-BOARD_VENDORIMAGE_FILE_SYSTEM_TYPE)
+else
+BUILD_VENDORIMAGE := $(shell CALLED_FROM_SETUP=true build/soong/soong_ui.bash --dumpvar-mode BOARD_VENDORIMAGE_FILE_SYSTEM_TYPE)
+endif
 endif
```
jamuir: I think this breaks Nougat builds, which are still supported.

Maybe do this instead:

```make
ifneq ($(filter 28, $(call get-allowed-api-levels)),)
# new cmd for pie and newer releases
else
# old cmd
endif
```

jamuir: thanks for your update.  It looks good.

just curious -- is there a reason to prefer

```make
ifeq ($(shell test $(PLATFORM_SDK_VERSION) -gt 27; echo $$?),0)
```
over
```make
ifneq ($(filter 28, $(call get-allowed-api-levels)),)
```
 ?

Stricted: since i have no clue what get-allowed-api-levels does or how it works i preffered to use something i know

jamuir: https://github.com/opengapps/aosp_build/blob/9867b7268ee44525f15ffd84cd84ee1521b3a77f/core/definitions.mk#L9-L11

`get-allowed-api-levels` is used quite a lot in aosp_build.

Stricted: updated my commit accordingly

jamuir:
thanks!

PR #225, opened 2019-09-10
[WIP] Include webview overlay when forcing Chrome browser.

MarijnS95:
Chrome supplies a webview as well, but is not listed as such when GAPPS_FORCE_WEBVIEW_OVERRIDES is not true. It has to be listed in the webviewproviders overlay to show up.
Subsequently, setupwizard fails halfway when it needs to use a webview.

This is probably wrong as it is overriding webviews without `GAPPS_FORCE_WEBVIEW_OVERRIDES` set. Can we come up with an alternative that makes Chrome the sole webview provider without including `WebViewChrome` in the build?


MarijnS95:
Closing in favor of https://github.com/opengapps/aosp_build/pull/234

Chrome on API/SDK level 29 does not provide a WebView anymore.

PR #227, opened 2019-11-05
Update README.md

nezorflame:
Add info regarding new project repos and move to LFS


mfonville:
I think AOSP also needs some extra work on the scripts used, to drop the lzip trick

PR #230, opened 2019-11-29
find_apk: Do not use ls -v anymore

dasJ:
Android 10 (or at least LineageOS 17) uses a prebuilt ls from
prebuilts/build-tools/path/linux-x86/ls. This ls does not have support
for -v, so this is what I came up with.


jamuir:
thanks for your PR!

I think we hit an issue like this previously and our quick and dirty solution was just to reference the system's `ls` explicitly:  i.e. change `ls -v` to `/bin/ls -v`.

jamuir:
your change looks good.  merging.

dasJ:
Thanks! I might have some more in the pipeline

PR #234, opened 2019-12-10
[Android 10] Add Trichrome WebView providers

MarijnS95:
The first commit adds the new WebView providers that depend on a secondary Trichrome package, which is shared between the WebView and Chrome itself. These are used on Android 10 instead of bundling the WebView directly into the Chrome package.

The second commit moves this overlay with the `webviewproviders` to an RRO named `GoogleWebViewOverlay.apk` just like Google does on their own releases. This is cleaner than bundling the overlay with the device-specific one, and should prevent it from missing if ever building a GSI with `framework-res` as enforced RRO (ending up on the vendor partition, not in the GSI).


MarijnS95:
I've pushed another commit that adds the TrichromeLibrary powering Chrome and WebViewChrome in Android 10. Without this, both refuse to run for not finding their `libmonochrome.so` which only resides in `com.google.android.trichromelibrary`.

This file still needs to be added to the scripts in the main repository, as well as the (`ten` branches in the) source repositories.

Sorry for tagging you @nezorflame, but you're working on Android 10 compatibility: how is this issue solved in the betas? I couldn't find trichrome in the repositories nor the beta releases, even though the same Chrome and WebViewChrome require it here. There doesn't seem to be another package that provides `libmonochrome.so` either.
(I haven't flashed the package on a gapps-less Android 10 ROM yet, does it even work there or fall back to system webview and require a manual Chrome/other browser install?)

MarijnS95:
@alviteri I appreciate you merging this straight away, but keep in mind that `com.google.android.trichromelibrary` has _not_ been added to the sources yet (unless that process is ongoing now), which should break the build. Did you intend to merge this already?

Abhi011999:
Annotation 2019-12-13 030206

Can confirm it breaks the build, should i just revert last two commits to build successfully ?

MarijnS95:
@Abhi011999 Apologies for the breakage, this was not supposed to be merged just yet. You only need to revert the last commit that adds the `TrichromeLibrary` package.

Keep in mind that this _should_ result in a broken WebViewChrome/Chrome, so you might want to put a trichrome package in the right repo instead.

alviteri:
@MarijnS95 just send the pr to other sources then it will be merge

MarijnS95:
@alviteri Shouldn't this be added in one of the scripts to get automatically updated?

chirayudesai:
FYI https://groups.google.com/a/chromium.org/forum/#!topic/android-webview-dev/Y8FIRqJCCNA

Chromium 80 and above should include Trichrome and webview for Q in the source code, 79 and below won't work with Q

nezorflame:
Thanks for the PR!
Yeah, I'm pretty sure this is required and Chrome won't work without this in the Q builds. I didn't check all of the apps yet, so it's probably broken right now.
As for the lib, just add it with PR in the source for now.
@NicholasBuse is the one managing the updates, tagging him so he can add this lib to the scripts.

MarijnS95:
@nezorflame Looking at https://github.com/opengapps/apkcrawler PR'ing to the sources is the way to get it emitted from `./report_sources.sh` and downloaded, assuming the automatic update scripts are set up exactly like the examples in the readme.

However, in order to PR I will need to make an account on the opengapps gitlab instance, fork the repo, push the apk and create a merge request. It's easier for someone with write access to push it directly, and integrate it properly with the ignore/blacklists in the crawler scripts.

Issue #237, opened 2019-12-16
Build failure - TrichromeLibrary: No source files specified


Viv3k:
Hi,

I'm getting the below error while I'm trying to build aosp with opengapps. I see that there are some latest changes around the same TrichromeLlibrary (ref: [commit](https://github.com/opengapps/aosp_build/pull/234)). Please let me know if there is anything that needs to be done to overcome this error.

I'm currently builing aosp  `android-10.0.0_r15` with `stock` variant of opengapps for walleye. @MarijnS95 Please help here if possible.

#### Error:
```
zipinfo:  cannot find zipfile directory in one of vendor/opengapps/sources/arm64/app/com.google.android.apps.translate/23/nodpi/64000865.apk or
          vendor/opengapps/sources/arm64/app/com.google.android.apps.translate/23/nodpi/64000865.apk.zip, and cannot find vendor/opengapps/sources/arm64/app/com.google.android.apps.translate/23/nodpi/64000865.apk.ZIP, period.
[ 97% 510/521] including vendor/opengapps/build/modules/TrichromeLibrary/Android.mk ...
FAILED:
vendor/opengapps/build/modules/GoogleWebViewOverlay/Android.mk: error: TrichromeLibrary: No source files specified
build/make/core/prebuilt_internal.mk:37: error: done.
20:02:25 ckati failed with: exit status 1

#### failed to build some targets (04:09 (mm:ss)) ####
```

Viv3k:
For anyone else facing the issue, you might want to revert the last commit on `master`.

https://github.com/opengapps/aosp_build/pull/234#issuecomment-565167472

https://github.com/opengapps/aosp_build/pull/234#issuecomment-565197357

MarijnS95:
@Viv3k Also apologies to you for breaking the build. That commit (entire Pull Request) should not have been merged until _after_ the dependencies were added to the relevant (auto) fetch scripts and added to the build scripts.

Though, without this commit your build should have a broken Chrome/WebViewChrome until it is updated from the Play store (or another browser/webview is used) or `com.google.android.trichrome` is manually pushed to the system.

@NicholasBuse Tagging you again for support with the crawlers/fetch scripts: what should be done to get TrichromeLibrary added to the source repos and automatically updated?

NicholasBuse:
What needs to get added to the source repo(s)?  .APK, .SO? Which repo (all, arm, etc.) what path?
[...]

PR #243, opened 2020-01-25
modules: Load TrichromeLibrary module only on API >= 29.

MarijnS95:
Fixes https://github.com/opengapps/aosp_build/issues/237

The APK for this module only exists on API level 29. Builds for Android
9 and below will fail due to the missing APK on a lower or matching API
level.
Besides, this module is never added to LOCAL_REQUIRED_MODULES on API
levels lower than 29 anyway.

Fixes: 191c464a3fcb28b1938a47ba74713c712101c3cd
Signed-off-by: MarijnS95 <marijns95@gmail.com>


nexus511:
@nezorflame Is it possible to merge this PR?

This fixes the build for Android 9.

PR #247, opened 2020-02-03
Build: Filter out opengapps-permissions based on os version

AnggaSP:
Currently, the builder will copy both permission into /system/etc/default-permissions which may cause problem for older Android version. We should conditionally copy the permission based on the os version.


nexus511:
@nezorflame Can you please merge this PR? This fixes the bootloop issue for Android 9. This is related to opengapps/opengapps#799

PR #260, opened 2020-06-05
Add tvgapps support

webgeek1234:
This also has a tvmini variant. I've had patches for that for a while, but never PRed it to the main repo since it doesn't help enough on the space constrained devices.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -21,10 +21,16 @@ endif
 GAPPS_VARIANT_EVAL := $(call get-gapps-variant,$(GAPPS_VARIANT))
 
 ifeq ($(GAPPS_VARIANT_EVAL),)
-  $(error GAPPS_VARIANT $(GAPPS_VARIANT) was not found. Use of one of pico,nano,micro,mini,full,stock,super)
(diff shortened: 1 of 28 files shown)


nezorflame:
@webgeek1234 great job!
@MarijnS95 could you take a look?

nezorflame:
Also @mfonville - is it OK to add another variant aka `tvmini`?

webgeek1234:
https://github.com/webgeek1234/opengapps/commit/402a30dd8d26d4554816f7f9f7eff905431ec33d
For reference, this is what I've got for the main repo. I can rebase and PR that too if desired.

mfonville:
Adding tvmini for the main repo would be great too, so that we can keep it in sync. If you can PR it, please!

MarijnS95:
@nezorflame I'm late to the party but it all looks good from my side. Normal builds seem to work just fine too but I didn't do any rigorous testing here.

That said we should at some point convert the actual scripts to soong bootstrap modules and the rest to blueprint, as this increasing number of makefiles slows down the (already massively slow) `mk` parsing stage. Blueprint is a lot less flexible allowing much faster parsing while the scripts themselves are compiled Go and can run at "native" speeds. Only for more recent Android versions though :(

PR #272, opened 2020-09-22
Initial support for Android 11

MarijnS95:
Please review individual commits for implementation details. Obviously requires Android 11 (SDK 30) apks of some apps (setupwizard, gms, gsf, (tri)chrome, webview, etc) to run on-device.

~This requires testing on Android 10 before it can be merged.~ Tested on Android 10.

@nezorflame any plans towards supporting `.apk.gz`, found on Android 11 Pixel images?

PR #277, opened 2020-10-14
[master] Satisfy check_elf_file.py requirements

pablomh:
If not, they will fail building (on Android 11, at least).

This has been tested on top of https://github.com/opengapps/aosp_build/pull/272.


MarijnS95:
Following our discussion in https://github.com/MarijnS95/opengapps_aosp_build/pull/1#issuecomment-708345327 we have decided to specify dependencies explicitly instead of disabling this check, in hopes of getting alerted about potential missing dependencies in the future during the build rather than when deploying on-device.

Issue #253, opened 2020-04-03
ActionsServices: No source files specified error


Road-Drum:
Following the [Getting Started](https://github.com/opengapps/aosp_build#getting-started), during building I got the following error:

```
[ 88% 573/651] including vendor/opengapps/build/modules/ActionsServices/Android.mk ...
FAILED: 
zipinfo:  cannot find or open , .zip or .ZIP.
tools/tradefederation/core/util-apps/WifiUtil/Android.mk: error: ActionsServices: No source files specified 
build/make/core/prebuilt_internal.mk:37: error: done.
10:04:10 ckati failed with: exit status 1
```

The build succeeds prior to adding opengapps manifest. What is actually missing in my case? seems that this issue is somewhat identical to #231 but I am building for `arm` target, not `x86`

stefanhh0:
Hello,

Since 05-Nov-2019 you have to use git-lfs, see also the README.md.

To actually fill the directory with the apk-files use the command:
`git lfs pull`

Best regards,
Stefan

Road-Drum:
Thanks, I've added `git lfs` now and `git lfs pull` but got this error instead

```
Error updating the git index: (235/235), 1.1 GB | 12 MB/s                                                                                                                                                     
error: app/com.google.android.gm/21/nodpi/62133988.apk: cannot add to the index - missing --add option?
fatal: Unable to process path app/com.google.android.gm/21/nodpi/62133988.apk
Errors logged to /home/hello/rpi-10/vendor/opengapps/sources/all/.git/lfs/logs/20200403T131053.217603936.log
Use `git lfs logs last` to view the log.
```

never used `git lfs` before. sorry about that

stefanhh0:
Not so sure why it can't be processed properly. 
- Have you retried? 
- Some file permission problem? Are you owning all the files?
- If nothing helps, remove the repo completely (also in `.repo/projects/vendor/opengapps`) and recreate accordingly.

Road-Drum:
>     * Have you retried?
> 
>     * Some file permission problem? Are you owning all the files?
> 
>     * If nothing helps, remove the repo completely (also in `.repo/projects/vendor/opengapps`) and recreate it accordingly.

I've retried on `vendor/opengapps/sources/all` and `vendor/opengapps/sources/arm`. the `arm` passed `git lfs pull` on first try but not `all`. tried again on `all` and the error doesn't pop up anymore.

[...]

PR #278, opened 2020-10-22
ignore TrichromeLibrary since it doesn't extist for arm

root-intruder:
similar to issue #222 I get build errors during makefile gathering for nonexisting packets on arm:

``[ 99% 13085/13114] including vendor/opengapps/build/modules/TrichromeLibrary/Android.mk ...                                             
FAILED:                                                                                                                                 
zipinfo:  cannot find or open , .zip or .ZIP.                                                                                           
vendor/opengapps/build/modules/GoogleWebViewOverlay/Android.mk: error: TrichromeLibraryGoogle: No source files specified                
build/make/core/prebuilt_internal.mk:37: error: done.                                                                                   
10:28:56 ckati failed with: exit status 1  ``

...since TrichromeLibrary doesn't exist on arm only on arm64...thus it should be excluded from makefiles

### DIFF
--- a/modules/TrichromeLibrary/Android.mk
+++ b/modules/TrichromeLibrary/Android.mk
@@ -1,4 +1,5 @@
 ifneq ($(filter 29,$(call get-allowed-api-levels)),)
+ifneq ($(TARGET_ARCH),arm)
 
 LOCAL_PATH := .
 include $(CLEAR_VARS)
@@ -7,5 +8,5 @@ LOCAL_MODULE := TrichromeLibraryGoogle
 LOCAL_PACKAGE_NAME := com.google.android.trichromelibrary
 
 include $(BUILD_GAPPS_PREBUILT_APK)
-
+endif #arm
 endif # API >= 29


MarijnS95:
@root-intruder Do you happen to know if Google provides a Chrome/webview variant of arm that depends on trichrome, but has simply not been added to the OpenGapps repos yet?

root-intruder:
no I haven't researched much but according to https://github.com/opengapps/aosp_build/issues/253#issuecomment-608975951 there should be...but since I don't want webview etc. in the rom I didn't further research

Blystad:
There exists a TrichromeLibrary for arm and arm64. https://www.apkmirror.com/apk/google-inc/trichrome-library/trichrome-library-86-0-4240-110-release/

I guess this should be added to OpenGapps @mfonville

mfonville:
@nezorflame and @NicholasBuse what is your take on this?

nezorflame:
While it does exist, it never worked for me when I was testing it. Need to take a second look at it though, maybe something changed.

PR #283, opened 2020-12-08
Add AndroidAuto module and add to stock target

Tortel:
There is currently no AndroidAuto module, so add it and add it to the stock target to bring inline with the zips.

Note that there are some privapp-permission updates needed for Auto, at least on 11 - this patch is required: https://gist.github.com/Tortel/a3d952eeea23dc13447376319e6dbb90

I created an account on GitLab to open a pull request for that change, but my account is pending approval.


alviteri:
let's see https://gitlab.opengapps.org/opengapps/all/-/merge_requests/43

